PR #2, opened 2013-11-19
Add link to Templates spec

zenorocha:


hayatoito:
The template spec has graduated and has been merged to HTML living standard.

http://www.whatwg.org/specs/web-apps/current-work/multipage/scripting-1.html#the-template-element

zenorocha:
So are you guys going to remove it from this repo?

hayatoito:
Yes. That should be removed to avoid a confusion.

zenorocha:
But isn't Templates "part" of Web Components?

hayatoito:
No longer, IMO. I guess there are outdated documents.
They should link to http://www.whatwg.org/specs/web-apps/current-work/multipage/scripting-1.html#the-template-element.

hayatoito:
Anyway, thank you for your attention! We have a lot of things to be cleaned up. :)

PR #5, opened 2014-03-17
BUG: Lifecycle code sample contains unexpected ":"

MarkBennett:
It looks like this sample was converted from using object literal notation to use function expressions, but one of the ":" was missed. I've replaced it with an assignment so that the sample works again.


MarkBennett:
This resolves the bug https://www.w3.org/Bugs/Public/show_bug.cgi?id=24398

dglazkov:
Thanks!

PR #6, opened 2014-04-07
Optimized images

j9t:
Noticed images can be further compressed; optimized via ImageOptim.


hayatoito:
Hmm... I don't think this kind of optimization is worth while here. Our workflow of updating the specs doesn't contain such an image compression task. I'm afraid that if we update these images in the future, your optimization will be taken away. We don't have a plan to have such an optimization task in our workflow.

j9t:
Even if it’s not feasible for the team to integrate image compression in their workflow (it’s easy to automate, and even ImageOptim only means a drag and drop), isn’t rejecting this change like walking away from a free oil change? It’s still an improvement.

(Not going to make this my concern, but neither the rationale for not compressing nor for rejecting this change list seem strong to me.)

hayatoito:
Yeah, and I feel sorry that we couldn't accept your optimization. However, I still have a concern that your optimization would be overridden in the future by us accidentally. I'm not a fan of such a situation.

I've taken a closer look your changes again. For images which are used by published working drafts, it might be worth while to accept this kind of optimization. However, I think the working drafts are served from the W3C's site, not from here. The assets in the /publish directory were just used to publish WD. They aren't being directly referred to from the world.

What do you think about sending you optimization to W3C? I'm afraid that It's too late to update the assets in /publish directory here if our goal is to reduce the size of images which are currently used in published WD.

j9t:
No, I think I’m out. I love to chime in here and there but like many others I’m doing this in my infinite free time. The only recommendation I can give, as a professional, is to put image optimization on your roadmap.

Cheers!

PR #7, opened 2014-04-18
Removed reference to bug marked as fixed.

KenjiBaheux:
Do we still need the reference to bug 23710 (marked "for later")?


KenjiBaheux:
@omo Minor edit to remove reference to a bug marked as fixed (Document vs DocumentFragment). Can you merge this?

PR #9, opened 2014-05-05
[Custom]: Missing created callback should not prevent other callbacks

dglazkov:
Missing created callback should not prevent other callbacks from being enqueued.

https://www.w3.org/Bugs/Public/show_bug.cgi?id=25142


dglazkov:
@coonsta PTAL?

PR #12, opened 2014-05-09
[Custom]: No way to associate class/constructor function.

dglazkov:
https://www.w3.org/Bugs/Public/show_bug.cgi?id=24019

### DIFF
--- a/spec/custom/index.html
+++ b/spec/custom/index.html
@@ -83,6 +83,7 @@
     </ol>
     </li>
     <li><a href="#parsing">Parsing Custom Elements</a></li>
+    <li><a href="#es6">Custom Elements and ECMAScript 6</a></li>
     <li><a href="#appendix-a">Appendix A: All Algorithms in One Diagram</a></li>
     <li><a href="#acknowledgements" class="no-number">Acknowledgements</a></li>
 </ol>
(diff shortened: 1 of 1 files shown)


dglazkov:
@arv PTAL? Setting up the basics first.

spec/custom/index.html
```diff
@@ -632,6 +633,81 @@ <h2 id="parsing">Parsing Custom Elements</h2>
...
+<dl>
+<dt>Input</dt>
+    <dd><var id="var-options-prototype">FUNCTION</var>, the custom element constructor function.</dd>
+    <dd><var>DOCUMENT</var>, the owner <a href="http://dom.spec.whatwg.org/#concept-document">document</a> for new <a href="#dfn-custom-element">custom element</a></dd>
+<dt>Output</dt>
+    <dd><var>FUNCTION</var>, the mutated custom element constructor function</dd>
+</dl>
+<ol>
+    <li>If <var>FUNCTION</var> is already an <a href="http://www.w3.org/TR/WebIDL/#dfn-interface-object">interface object</a> for any <a href="http://www.w3.org/TR/WebIDL/#dfn-interface">interface</a>, throw a <a href="http://dom.spec.whatwg.org/#notsupportederror"><code>NotSupportedError</code></a> and <strong>stop</strong>.</li>
+    <li>Let <var>DEFINITION</var> be an <a href="#dfn-element-definition">element definition</a> that has <var>FUNCTION</var> as <a href="#dfn-custom-element-prototype">custom element prototype</a></li>
```
arv, line 694: This is weird. FUNCITON is not the prototype. Maybe we need to use other terminology?

arv, line 694: Should probably be "custom element constructor function" instead?

arv:
LGTM

spec/custom/index.html
dglazkov, line 694: Right. This will be broken until I fix https://www.w3.org/Bugs/Public/show_bug.cgi?id=24020.

PR #13, opened 2014-05-12
Use attribute local name, rather than attribute name.

dglazkov:
@annevk Is this better?


hayatoito:
Could we close this pull request?

PR #17, opened 2014-06-10
Qualify "this specification" with "this version"

AFBarstow:
As discussed on public-webapps, the text about the spec being obsoleted needs to be qualified with "this version". See http://lists.w3.org/Archives/Public/public-webapps/2014AprJun/0825.html.


hayatoito:
Thank you!

PR #32, opened 2015-01-19
[Shadow DOM] Add Attributes section

kojiishi:
PTAL!


hayatoito:
LGTM. Let's add this section until we can upstream these into HTML Living Standard.

PR #35, opened 2015-02-09
Describe focus navigation order for the case with tabindex=-1

TakayoshiKochi:
This is a proposed draft of shadow navigation order when shadow host or <shadow> with
tabindex=-1, corresponds to https://www.w3.org/Bugs/Public/show_bug.cgi?id=27965 .


hayatoito:
Thanks. Before accepting this PR, could you drive a discussion in #86?
We need an agreement since this is backward-incompatibility change.

hayatoito:
I believe that this PR is obsolete since Blink is now experimenting a delegatesFocus. Let me close this.

PR #36, opened 2015-03-03
add autolinking as a preprocessor

darobin:


hayatoito:
LGTM. Thank you for fixing!

PR #41, opened 2015-04-09
Add "http:" to the respec URL so that respec runs when viewing locally

kojiishi:
This patch allows re-spec to run locally, which is helpful when editing.
If this was intentional for some reasons, I can keep this locally, but wanted to ask if so.
Note that the [ReSpec template](http://www.w3.org/respec/examples/template.html) does have "http:".

### DIFF
--- a/spec/shadow/index.html
+++ b/spec/shadow/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+﻿<!DOCTYPE html>
 <html>
   <head>
     <meta charset='utf-8'/>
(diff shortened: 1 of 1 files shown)


sideshowbarker:
I believe the lack of the `http:` here is intentional—so that the the content can be served both from insecure http and also from https without generating mixed-content warning indicators.

hayatoito:
Yeah, it's intentional. If we add "http:", UA would warn "the resource is insecure" if index.html itself is served by "https:".
You should use local web server to run locally.

PR #42, opened 2015-04-17
Change the order of insertion points which are involved in a re-distribution in event path (W3C bug 23887)

kojiishi:
As the result of discussion in [W3C bug 23887](https://www.w3.org/Bugs/Public/show_bug.cgi?id=23887), I'm making a PR of the algorithm change to what so-called "candidate 2".

Both the algorithm and the example result changes are included in this PR. Charts are not updated but marked as an issue.


hayatoito:
Let me close this PR since this is obsolete.

PR #45, opened 2015-05-01
Make proposals more clear

annevk:
I added IDL to 1. Made 2 significantly simpler to reflect discussion on the list. No need to keep insertionList thingie alive based on the feedback given. If someone wants to champion that still they can make it a new proposal. Also added alternatives 3/4 to the triggering API proposals.


annevk:
You can view this here: https://github.com/annevk/webcomponents/blob/patch-1/proposals/Imperative-API-for-Node-Distribution-in-Shadow-DOM.md

PR #47, opened 2015-05-01
Added pure functions idea to the timing alternatives.

dglazkov:


dglazkov:
@rniwa @annevk PTAL.

proposals/Imperative-API-for-Node-Distribution-in-Shadow-DOM.md
```diff
@@ -210,6 +210,18 @@ Introduce a variant of mutation observers where the callback is invoked after ea
 - It could be as messy as old mutation events
 - Might be still too expensive to use with the second approach which requires collecting every distribution candidate.
 
-## 4. Not a problem for v1
+## 5. Run Distribution Callback in a Separate Scripting Context or as Pure Function
```
rniwa: Perhaps this should be 4?

proposals/Imperative-API-for-Node-Distribution-in-Shadow-DOM.md
```diff
@@ -210,6 +210,18 @@ Introduce a variant of mutation observers where the callback is invoked after ea
...
+
+### Pros
+- Provides consisten distribution state to user code
+- Interoperable
+
+### Cons
+- Both separate scripting context and pure function ideas need much more thorough examination and may take years to get right
+
+
+## 4. Not a Problem for v1
```
rniwa: and this be 5?

dglazkov:
Whoops, done.

annevk:
This seems like a mix between timing and API alternative, but I guess it's okay to put it in the timing bucket.

More importantly, do you really think we can define how pure functions behave, get agreement on that, and implement that within a year? And then not get complaints about all kinds of things not being available within the pure function environment?

From all ideas this really seems pie in the sky.

To be clear, we were hoping to have a complete plan for shipping Shadow DOM after the last meeting... July is actually pushing it for us, but it's starting to sound like we'll need to find some alternative plan that does not involve shipping it.

PR #53, opened 2015-05-21
Added a couple of algorithms and definitions.

dglazkov:
Started fleshing things out in the proposal. @rniwa @annevk @hayatoito PTAL.


hayatoito:
I think we have to think how the current "content" and "slot" can interact each other if both are used together.
In Blink, we will have a painful migration period where both "content" and "slot" co-exist.

hayatoito:
The easiest way is:
If a tree of trees has at least one "slot" element, "content" element will be no-op.

PR #55, opened 2015-05-22
id -> name

annevk:
We should consistently use name I think. If they really have to be unique, and I guess they have to be, maybe we should simply reuse the id attribute? Or would that create a different set of problems?


hayatoito:
lgtm

Issue #61, opened 2015-05-25
[Shadow]: Update constraints around stopping events (bugzilla: 20247)


hayatoito:
Title: [Shadow]: Update constraints around stopping events (bugzilla: 20247)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=20247

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=20247#c0
_Dimitri Glazkov_ wrote on 2012-12-05 15:27:41 +0000.

"The following events must always be stopped at the nearest shadow boundary:"

Since we relaxed the lower boundary in regard to events, we should similarly not stop events there.

---

comment: 1
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=20247#c1
_Dimitri Glazkov_ wrote on 2012-12-05 15:31:06 +0000.

Also, this needs to account for the nodes, distributed to the insertion point. Events from those should be heard outside of the shadow tree.

---

comment: 2
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=20247#c2
_Dimitri Glazkov_ wrote on 2012-12-05 17:54:57 +0000.

And we need to write down the rationale behind stopping these events.

---

comment: 3
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=20247#c3
_Dimitri Glazkov_ wrote on 2012-12-06 23:08:37 +0000.

I am trying to come up with a general rule of thumb. I guess is something like:

If the event is retargeted, does event's meaning change?

For example, if a "load" event is heard at a shadow host, it conveys the meaning of shadow host somehow completing loading, which is not true, because all it means that some element inside of host's shadow tree has completed loading.

Conversely, if an "click" event is heard a a shadow host, it says that the shadow host was clicked. Since shadow tree is how shadow host is rendered, the meaning of the event stays the same.

WDYT?

---

comment: 4
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=20247#c4
_Anne_ wrote on 2012-12-07 10:50:53 +0000.

But if the shadow tree actually did load something via XMLHttpRequest and wanted to convey that you're lost.

I suppose the current set of events is based on what WebKit needed? It seems unclear that will match the needs of developers when this gets out in the wild.

---

comment: 5
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=20247#c5
[...]

PR #274, opened 2015-07-07
Changes to Dispatching Events

kojiishi:
@hayatoito, @annevk, could you review? This is for issue #61.

### DIFF
--- a/spec/shadow/autolink-config.js
+++ b/spec/shadow/autolink-config.js
@@ -17,6 +17,7 @@ var autolinkConfig = {
         'document': '#concept-document',
         'document element': '#document-element',
         'element': '#concept-element',
+        'event constructor': '#concept-event-constructor',
         'event dispatch': '#concept-event-dispatch',
         'event listener invoke': '#concept-event-listener-invoke',
         'event listener': '#concept-event-listener',
--- a/spec/shadow/index.html
+++ b/spec/shadow/index.html
@@ -510,15 +510,17 @@
 
       <section>
         <h3>Changes to Dispatching Events</h3>
+        <p>Insert the following item after item 3 of <a title="event constructor">Constructing events</a>:
+        <ol start="4">
+          <li><ins>Initialize event's <a>scoped flag</a> as defined in
+            Events that are not leaked into ancestor trees.</ins>
+        </ol>
         <p>The item 4 of <a title="event dispatch">Dispatching events</a> is changed to:
         <ol start="4">
-          <li>If event’s target attribute value is participating in a tree,
-            <ins>initialize event's <a>scoped flag</a> as defined in
-            Events that are not leaked into ancestor trees, and</ins>
-            let event path be
-            <ins>the output of the <a>event path calculation algorithm</a> with target and event as inputs.</ins>
-            <del>a static ordered list of all its ancestors in tree order,</del>
-            <ins>Otherwise clear event's <a>scoped flag</a></ins> and let event path be the empty list<del> otherwise</del>.
+          <li>If event’s target attribute value is participating in a tree, let event path be
+            <ins>the output of the <a>event path calculation algorithm</a> with target and event as inputs</ins>
+            <del>a static ordered list of all its ancestors in tree order</del>,
+            and let event path be the empty list otherwise.
         </ol>
       </section>
 
--- a/spec/shadow/index.html
+++ b/spec/shadow/index.html
@@ -514,9 +514,9 @@
         <ol start="4">
           <li>If event’s target attribute value is participating in a tree,
             let event path <ins>and the <a>scoped flag</a></ins> be
-            <ins>the result of the <a>event path calculation algorithm</a>.</ins>
+            <ins>the output of the <a>event path calculation algorithm</a> with target and event as inputs.</ins>
             <del>a static ordered list of all its ancestors in tree order,</del>
-            <ins>Otherwise L</ins><del>and l</del>et event path be the empty list <del>otherwise</del>
+            <ins>Otherwise</ins><del>and</del> let event path be the empty list <del>otherwise</del>
             <ins>and clear the <a>scoped flag</a></ins>.
         </ol>
       </section>
--- a/spec/shadow/index.html
+++ b/spec/shadow/index.html
@@ -513,11 +513,12 @@
         <p>The item 4 of <a title="event dispatch">Dispatching events</a> is changed to:
         <ol start="4">
           <li>If event’s target attribute value is participating in a tree,
-            let event path <ins>and the <a>scoped flag</a></ins> be
+            <ins>initialize event's <a>scoped flag</a> as defined in
+            Events that are not leaked into ancestor trees, and</ins>
+            let event path be
             <ins>the output of the <a>event path calculation algorithm</a> with target and event as inputs.</ins>
             <del>a static ordered list of all its ancestors in tree order,</del>
-            <ins>Otherwise</ins><del>and</del> let event path be the empty list <del>otherwise</del>
-            <ins>and clear the <a>scoped flag</a></ins>.
+            <ins>Otherwise clear event's <a>scoped flag</a></ins> and let event path be the empty list<del> otherwise</del>.
         </ol>
       </section>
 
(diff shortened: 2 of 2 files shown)


kojiishi:
> You also need to pass the scoped flag in somehow, no?

The algorithm receives `node` and `event` as inputs, `path` as output. The scoped flag is set to the `event`.

> Lowercase "l".

Done.

> That algorithm would need some arguments I think, otherwise we don't get a useful result.

Clarified that `node` and `event` as inputs.

I'm a bit worried if I understand your reviews correctly, please let me know if I don't look to understand correctly.

kojiishi:
Oh, wait, I think I understand what I misunderstand. I'll prepare another one.

annevk:
Ah yes, `scoped flag` is part of the event, that makes sense I suppose. In that case though you need to talk about "_event_'s scoped flag" though. It cannot just appear out of nowhere. And it cannot be the output of some algorithm in that case.

kojiishi:
Right, so initialize the `scoped flag` and then give the `event` as input. Does this look good?

annevk:
I see, it seems that the "scoped flag" should not be mentioned here at all. That is for the specifications that define and dispatch the event to set. E.g.  would set that flag for its events. We could probably define a convenient "fire a simple scoped event" shortcut for those specifications or some such.

kojiishi:
Strictly speaking, we need to branch after item 3 of this section because `target` could be overridden in item 3, and the `scoped flag` depends on whether the `target` is in-document or not.

I went though HTML spec, finding that the `target override` when used is always the `Document` object, so we can ignore this minor details.

Another option is to set the `scoped flag` regardless of in-document or not. We can then initialize the flag in [3.4. Constructing events](https://dom.spec.whatwg.org/#constructing-events). It only affects the event path, which is empty anyway if the `target` is not in-document, so the flag does not matter at all for XMLHttpRequest.

This might look cleaner. WDYT?

kojiishi:
I can't read whether the [initialize an event](https://dom.spec.whatwg.org/#concept-event-initialize) implies a call to a [constructor](https://dom.spec.whatwg.org/#constructing-events), I suspect it does?

kojiishi:
Updated the PR accordingly, or on [rawgit](https://cdn.rawgit.com/kojiishi/webcomponents/64a94660cb61e23a76542244c726de17d6e8e613/spec/shadow/index.html#changes-to-dispatching-events) for review.

kojiishi:
Oh, I think I understand what you mean by "shortcut"...let me try.

kojiishi:
The PR updated, do I understand what you meant correctly?

annevk:
It seems you removed the change to dispatching events where it invokes this new event path construction algorithm along with some arguments. We should keep that.

I guess other than that we just need to find all the event dispatch sites that need to set this flag on the event object before dispatching it...

kojiishi:
> It seems you removed the change to dispatching events where it invokes this new event path construction algorithm along with some arguments. We should keep that.

Revived.

> I guess other than that we just need to find all the event dispatch sites that need to set this flag on the event object before dispatching it...

Do you want to find all of them and put all patches here, or can we live with listing event types as in this PR? I put "not limited to", so any new specs can define additional scoped events if it wants to.

annevk:
A list of events is wrong. E.g. while `` will dispatch scoped events, `XMLHttpRequest` might not, yet they share event types.

kojiishi:
The item 4 of Dispatch Events has precondition of:

> If event’s target attribute value is participating in a tree

so we're fine, no?

annevk:
Well, what if we have some new tree-based API where we do want to leak an event with a type that is already mentioned? It seems like an ugly hack to base this on event type (which should not be meaningful other than as API contract) rather than dispatch location.

kojiishi:
Could you explain what you're suggesting a little more? I don't understand what you want this to be. It's LS, we can fix if new tree-based API appears, no?

annevk:
I mean that the way this is structured is broken. If we want `` to only dispatch scoped events, we should write the specification for `` such. We shouldn't have a list of types elsewhere that hopefully catches all offenders (and will catch no new ones since those writing APIs have no idea what is going on).

annevk:
The way this works in JavaScript

``` js
ev = new Event("yay", {scoped: true})
element.dispatchEvent(ev)
```

is how the specifications should work as well. There should not be a magic list.

kojiishi:
> ...is how the specifications should work as well.

The list only sets default value, so your JS should work with the current spec with this PR. The target participates in tree, so it goes to calculate event path, and the event path calculation algorithm outputs a scoped list if the scoped flag is set.

> we want `` to only dispatch scoped events, we should write the specification for `` such

I'm lost what you are suggesting, sorry, can you explain? I searched for, for instance, where WHAT WG spec mentions `scroll` event, found that it says "it's defined in CSS OM", and CSS OM does not define that, so it looks like it's not defined anywhere. You're not suggesting Shadow DOM spec should have all such event firing definitions, right?

annevk:
Yeah, I don't think `scroll` is well defined at the moment :-( I'm saying that we should not have default values based on the event type. It's fine to have a list of event types for monkey patching, but we shouldn't implement it that way and it shouldn't be the end goal for specifying things.

E.g. if `` starts dispatching a new event `tralala` we wouldn't want to update shadow DOM (and perhaps create issues for some other element that doesn't want a scoped `tralala` event). `` should just say it's a scoped event.

kojiishi:
> E.g. if  starts dispatching a new event tralala we wouldn't want to update shadow DOM (and perhaps create issues for some other element that doesn't want a scoped tralala event).  should just say it's a scoped event.

I think I tried the wording to resolve the concern; i.e., it says "The following events are the scoped events, but **not limited to**", so the new spec simply can say "the new event `tralala` is a scoped event" without updating Shadow DOM spec (or DOM if this is upstreamed by then.)

If it doesn't read so, maybe my English problem, but I'm suspecting that I'm not understanding your intention very well rather than poor English in this case.

What I don't understand is, if we don't patch DOM spec, and don't list up event types to set the `scoped flag` to true by default, how do I specify which existing events are the scoped events?

annevk:
We would list all the places that need updating to dispatch scoped events. E.g. HTML's ``, `<video>`, and `<audio>` elements. At least, I imagine we wouldn't want `loadedmetadata` to leak out of a shadow tree by default? Or is there a reason for not leaking `load` and `error`, but leaking that one?

The whole point of this "scoped flag" exercise is to get rid of this list of event types and instead update specifications to dispatch scoped events where needed. As long as specifications are not updated we can "monkey patch" and file bugs on them.

kojiishi:
I think I understand what you mean, so you want to grep all "fire event" sentences in HTML/DOM spec, and have patches to every one of them in Shadow DOM spec? I still wonder I'm misunderstanding, since the idea doesn't look good to me.

I understand your argument that the same event type may be used as a scoped event or non-scoped event, but having default does not prevent doing that. Rather, patching hundreds of lines look unmaintainable to me.

What did I miss, or do I understand correctly and just disagree?

annevk:
You understand correctly. Perhaps all events need to be scoped, in which case we could simply update "fire event", but I doubt that.

kojiishi:
> Perhaps all events need to be scoped, in which case we could simply update "fire event", but I doubt that.

Looked around the current spec.
- Most events are not simply mentioned in other than [Index - Events](https://html.spec.whatwg.org/multipage/indices.html#events-2), which says "Fired".
- Others refer to [Firing a simple event named _e_](https://html.spec.whatwg.org/multipage/webappapis.html#fire-a-simple-event)

and "a simple event" implies "a trusted event" according to the current definition:

> Firing a simple event named e means that a trusted event with the name e, which does not bubble (except where otherwise stated) and is not cancelable (except where otherwise stated), and which uses the Event interface, must be created and dispatched at the given target.

So changing s/a trusted event/a trusted and scoped event/ seem to cover all what you need. Is this satisfactory?

annevk:
It depends on whether everything that uses "fire a simple event" needs to fire a scoped event. What rules do we use to determine that?

hayatoito:
It looks this PR is out of date. Let me close this.

PR #292, opened 2015-08-04
Added proposed changes to the cascade order for Shadow DOM.

ghost:
Add proposal from www-style on request from kochi.


hayatoito:
Thanks.

FYI. @tabatkins, this is the proposal which we agreed roughly on www-style.
AFAIR, I think Blink and Polymer are okay to this change basically. We have to measure how this change would affect Polymer, at least.

We want to implement this new rule in Blink tentatively asap because the current implementation is based on a rule which has inconsistency and has been hard to implement efficiently.

@annevk , @rniwa , if you have bandwidth to consider how cascade order for Shadow DOM should work, we'd appreciate if you can give us early feedback.

PR #295, opened 2015-08-18
Fix typo and add missing selectors

arthurevans:
There were a couple of typos in here (specifaction, selectores, :shadow), and I couldn't quite understand the second sentence in the note, so I took a stab at rephrasing. Also, I think :host and :host() should be mentioned here.

I realize this section will be changing, so feel free to disregard this PR if it's not useful.


hayatoito:
Thanks! LGTM.

PR #332, opened 2015-09-28
Draft WD to be published on or around 2015-Oct-08

AFBarstow:
These files are for a new Working Draft publication of Shadow DOM, currently set for publication on 08 October 2015.


hayatoito:
Thank you!

PR #336, opened 2015-10-06
clearer, more accessible code

chaals:
Using role="img" and aria-label on end nodes that don't contain text, (because proper SVG support is still rubbish in accessibility terms, although nodes that contain text are navigable by default) makes the whole thing a lot more navigable e.g. with a screenreader.

Using aria-hidden on the defs stops them being read at the beginning (another bug in browsers)

Adding aria tree navigation markup, and reordering the tree, would mean the navigation even makes sense, but I'll do that later.


hayatoito:
Thank you. However, these svg files have not been directly edited by human. We are using OmniGraffle software. The original files are assets/images/*.graffile. The svg files were the output of the export by OmniGraffle.

I'm afraid that your changes will be easily overridden unless the root cause is resolved. :(

chaals:
Yeah, I realise they come from Omnigraffle.

I don't really mind having the changes overwritten - at this stage I'm just trying to work up the accessibility because so far nobody seems to be able to do that properly. At least being able to describe the process of fixing the images would be useful, IMHO.

(When people want to work on the spec, they really will need something accessible. The alternative is to add a longdesc for each image - that might be easier for pragmatic reasons…)

hayatoito:
Thank you, however, I'm afraid that we can not have bandwidth to fix exported images manually in each export. It would be nice to have a reproducible reliable way to fix exported images if they are broken in terms of accessibility. Fixing them manually doesn't sound a scalable way as a user of the software.

Should we consider it as OmniGraffle's bug? If so, we should report this to those who made OmniGraffle.

chaals:
So the real question is what we can do in the workflow to improve the accessibility of the spec, including the images. Are there images that are unlikely to change?, like the composed tree? Maybe we can keep that and then make accessible versions of whatever other images you make, by hand-editing from that base...

Doing the first one was a few hours work and I expect it will end up being another day to get right. But copying that out to the other images we have now might be a pretty quick task, if you don't think they will change… or if it is important that people can look at them and understand what is going on, in case they want to comment on how the whole thing works.

("Take out the images" is the wrong answer, of course. It's great that they are actually there, because they really help…)

chaals:
I'll wait for an answer to the question of how to integrate accessibility, but feel free at that point to close out the Pull Request.
thanks :)

chaals:
Omnigraffle could do a lot better at supporting accessibility, which I guess we should report. But the problem isn't just Omnigraffle.

I agree that we don't want to manually edit images every time you export them - that's a crazy waste of time. I'm not sure how easily we can make an automatically reproducible repair though. Some of the fixes could be automated, but some of the stuff is likely to be difficult - especially the things I haven't fixed yet.

And I don't have omnigraffle so I am not sure how easy it is to change what it is doing in a way that is scalable for you to do in reality...

...hence the suggestion that I manually work up "shadow" versions of the images (in a parallel directory?), and rely on you to tell me which seem stable and which seem likely to change. I could always watch github to see which images actually _do_ change, right?

hayatoito:
I'm okay to merge this PR, given that images are not frequently updated. Thank you for your efforts. I appreciate that.

I don't think we should spend our time on having automatic repair process. That's not worth while.

I won't change my workflow. I'll use OmniGraffle when I need to update the images and export that to svg files. Please forgive me to overwrite your changes in the next export.

chaals:
Yeah, I don't expect you to change your workflow, and I understand that if you need to change an image anything I have done there will get overwritten. 

If you can tell me which images you think are not going to change soon then I can happily work on more accessible versions as I have time.

I guess most of the images are stable, so if I have a lot of spare time it would be useful to make all the versions now, and then update anything that actually changes. But it won't happen as fast as that.

Instead I propose to make a parallel directory with improved versions that can be used - I think that is the most useful way forward that lets us keep working without breaking stuff or trying to change our workflows…

hayatoito:
Thanks. You don't need to update svg files if they are not referred from the editor's draft. It's okay to remove unused svg files if that's distractive.

I don't think we need to have a parallel directory. Please feel free to update svg files directly. We can retrieve the previous svg files anytime from git history.

PR #339, opened 2015-10-07
shadow assets improved for a11y

chaals:
One SVG, only half-finished improving… (but better than the version
already merged)

Keeping in a parallel directory as protection against having it
overwritten (I hope)


hayatoito:
Instead of having a parallel directory, could you keep the updated SVG files in the same directory?
Please feel free to update SVG files directly in the directory so that we don't need to update HTML file.

We don't need to protect against having it overwritten because we'll overwrite it with a better (or latest) SVG file, always.

chaals:
Sorry, I will do. I somehow got out of synch yesterday, but I think this PR now does what we both want.

hayatoito:
Thanks. LGTM.

PR #340, opened 2015-10-11
accessibility enhancements

chaals:
Can be explored in good browsers by keyboard-on users, and gives
something useful for more or less all modern screenreader setups

### DIFF
--- a/assets/images/composed-tree-desc.html
+++ b/assets/images/composed-tree-desc.html
@@ -1,0 +1,56 @@
+<!DOCTYTPE html>
+<html><head>
+ <title>composed-tree.svg description</title>
+</head><body>
+<h1>Composed-Tree.svg</h1>
+<p>A graphical tree, representing some nodes in an HTML document's object model.</p>
+<p>The document root node - level 1 - is linked to two second level nodes:<p/>
+<ul>
+ <li>Level 2, node 1, has two children
+  <ul>
+   <li>Level 3 node 1, a shadow host, with two children
+    <ul>
+     <li>Level 4, node 1, with two children
+      <ul>
+       <li>Level 5, node 1, with no children</li>
+       <li>Level 5, node 2, with no children</li>
+      </ul></li>
+     <li>Level 4, node 2, with two children
+      <ul>
+       <li>Level 5, node 3, with no children</li>
+       <li>Level 5, node 4, with no children</li>
+      </ul></li>
+    </ul></li>
+   <li>Level 3, node 2 with no children</li>
+  </ul></li>
+ <li>Level 2, node 2, has two children
+  <ul>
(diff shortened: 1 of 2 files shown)


hayatoito:
Just in curious, how the composed-tree-desc.html is used?

chaals:
As the target of a longdesc, if you are including the SVG via an `img` element... (or stick the content in the "fallback" of an `object`).
I wrote it because I was building test cases. But given that SVG support is still a bit shaky in terms of accessibility, I think it is a pretty useful backup to have - longdesc is supported by every screenreader I know except VoiceOver (Apple has some objection to it because they think SVG can do better - which could happen in theory but isn't actually true in practice, even of Safari), and most browsers handle it or have simple extensions available to do so.

chaals:
One reason to do this is "something more or less useful" is about the best I think we can do, even with the round of improvements I am still working on...

hayatoito:
Thanks, however, I'm afraid that we can't guarantee that composed-tree-desc.html and composed-tree.svg are always synced. Suppose that I overwrite composed-tree.svg in the future, users would get the outdated description because I won't touch *-desc.html files when overwriting svg files.

chaals:
Sure. On the other hand, right now they get nothing at all. It seems on balance of probability that the description that is there will be useful - even if you mess with the image in some noticeable way.
It seems most users prefer a high probability of a useful description, with a risk of it being obsolete and therefore wrong, over zero probability of any description.
As a backup, if you _do_ change the image so much that the description is useless, you could always let me know, and if I don't update then unlink it...

hayatoito:
Thanks. I really appreciate your efforts, however, I won't change my workflow.
Thus, we can't guarantee that description stays useful. The description could be confusing to users, I'm afraid.

I agree that it would be nice to have a good description for svg files, but I don't have bandwidth to work on this. I'd like to work on other P1 spec issues, which should be resolved asap.

Could you only update *.svg files?

LJWatson:
@hayatoito
"I agree that it would be nice to have a good description for svg files..."

For those of us who can't see the SVG, the description is a bit more important than "nice to have".

The PR includes a description that is useful at present, and there is an offer to help keep the description up to date if that isn't something you can do.

Merging a PR, and tagging @Chaals in any commits where the SVG is updated, aren't things that should take any time or effort. They are things that will make the spec easier for blind people to use though, so I hope you'll reconsider.

hayatoito:
If we want to add a description for each SVG file and keep it up-to-date, we should have a git {pre or post}-commit hook which gives us a warning message when we update only a SVG file.

Unless such a warning message, we are likely to have the obsolete description files. We are humans.

hayatoito:
I've filed an issue: https://github.com/w3c/webcomponents/issues/341
Let's continue the discussion there and find the best way to improve the accessibility.

Issue #358, opened 2015-12-21
[Shadow] activeElement behavior seems to break encapsulation


bicknellr:
It seems strange for a ShadowRoot's `activeElement` to point to elements outside of that ShadowRoot. In particular, the spec currently seems to allow a ShadowRoot's `activeElement` to point not only to elements within the ShadowRoot but also elements in the tree that the ShadowRoot's host participates in (and possibly even further outwards).

I think it would be more reasonable if `activeElement` of a ShadowRoot could only point to an element contained within that ShadowRoot (or be null if none was focused). Additionally, if the 'deepest' focused node (i.e. the focused, UA-implemented element without a ShadowRoot) was nested further within descendant ShadowRoots, any given ShadowRoot's `activeElement` along that chain would point to the most shallow host of that chain contained within that ShadowRoot.

One implication of this approach is that determining focused descendants of the ShadowRoot's host means checking `activeElement` of the Document / ShadowRoot containing that host. This situation _seems_ to break encapsulation by implying that the code controlling the ShadowRoot / component needs to look at its surrounding tree. However, I would argue that it's preferable as descendants of the ShadowRoot's host (i.e. those outside of the ShadowRoot) should be the responsibility of the code controlling the tree in which those descendants (and the host) participate. Also, there's precedent for using a tree's root to determine focus within that tree given Document's `activeElement` property already works this way.

An option for providing the ShadowRoot with some insight into what non-shadow descendants of its host might be focused would be to have `activeElement` point at a slot when an element distributed to that slot (or descendant of) is focused. (Maybe even give slots `activeElement` that point to the focused non-shadow descendant? They seem a lot like ShadowRoot hosts with no descendants anyways.) I don't think this would really be necessary though, given that you could just check if the activeElement of the host's containing Document / ShadowRoot is contained within the host, but it might make working with focus of non-shadow tree descendants within a component easier.

In general, I think this would make the scope of responsibility of code watching focus of a component's nodes more closely mirror shadow tree boundaries.

(Also, I get the feeling that a ShadowRoot's encapsulation mode would somehow be relevant to this but I haven't been able to figure out anything about encapsulation modes from the spec other than that a ShadowRoot has one and it's either 'open' or 'closed'.)

hayatoito:
Thanks. Let me add an example, which came from our previous internal discussion, slightly modified so that it uses slot elements:

## Example

document tree: (Updated: I fixed all wrong indentation and missing closing tags)

``` html
<body>
[...]

PR #380, opened 2016-02-09
Proposed change for active element algorithm.

TakayoshiKochi:
Discussed in https://github.com/w3c/webcomponents/issues/358

### DIFF
--- a/spec/shadow/index.html
+++ b/spec/shadow/index.html
@@ -1157,10 +1157,30 @@
             <dd><var>ROOT</var>, either a <a>document</a> or a <a>shadow root</a></dd>
             <dd><var>ELEMENT</var>, the focused <a>element</a></dd>
             <dt>Output</dt>
-            <dd><var>ADJUSTED</var>, an adjusted <a>activeElement</a> property of <var>ROOT</var>.</dd>
+            <dd><var>ADJUSTED</var>, an adjusted <a>activeElement</a> property of <var>ROOT</var></dd>
           </dl>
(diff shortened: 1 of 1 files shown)


TakayoshiKochi:
@hayatoito PTAL

hayatoito:
Looks okay. However, the equivalent algorithm can be expressed by adding a few lines to the current algorithm:
1.   Let ADJUSTED be the result of the retargeting algorithm with ROOT and ELEMENT as input
2.   If ROOT is not the root node of ADJUSTED:
   1. Let ADJUSTED be null

I prefer this because this express the intention more clearly, IMO.

rniwa:
It seems premature to make this change given there is a disagreement on what `activeElement` should do.

TakayoshiKochi:
@hayatoito It looks good. I'm not sure if this algorithm having any dependency on changes in event retargeting algorithm, but as they are in the same spec it should not be a problem for any consistency.

@rniwa okay, will wait for other parties.

hayatoito:
I have named it "retargeting algorithm", instead of "event retargeting algorithm", because the concept of _retargeting_ is not for only for an event. This algorithm is intentionally written as a general algorithm so that we can re-use it from other places where we need a retargeting, such as activeElement here.

TakayoshiKochi:
@hayato got it.  The new patch has "retargeting algorithm", as you suggested.

hayatoito:
It looks outdated. Closing.

Issue #375, opened 2016-01-28
Focus Navigation for slots


rniwa:
We should update the [section 6.2 Focus Navigation](http://w3c.github.io/webcomponents/spec/shadow/#focus-navigation) to reflect the discussion we had during TPAC last year.

I think what we want is akin to putting nodes inside a shadow DOM and a slot into its own [nested browsing context](https://html.spec.whatwg.org/#nested-browsing-context).  We should probably edit the HTML spec to create some sort of scoping mechanism, and make shadow DOM / slot element / browsing context use that concept.

rniwa:
Could @hober or @annevk refactor the HTML spec as needed?

hayatoito:
I'm interested in how this new feature would interact with the proposal in https://github.com/w3c/webcomponents/issues/126.

If either of them is an backward incompatible feature, we can not ship Shadow DOM v1 unless both are resolved. 

@TakayoshiKochi
Could you lead the discussion? I'd like to know how we should handle the proposal in issue #126.

rniwa:
I don't think they're incompatible with one another since we still need to define the behavior when `delegateFocus` is missing.

hayatoito:
I guess @TakayoshiKochi has an opinion about this.
To me, they look apparently relevant if `deleteFocus` is enabled since either of them tries to define how tab focus navigation should work.

TakayoshiKochi:
The discussion at TPAC was that basically we navigate focusable elements via TAB presses in a way that those within the same shadow root including distributed nodes are grouped, and in order of composed tree?

e.g.

``` html
<my-nameinput>
    <#shadow-root>
         Firstname: <slot name="firstname"></slot>
         Lastname: <slot name="lastname"></slot>
    </#shadow-root>
    <input id="first" slot="firstname">
    <input id="last" slot="lastname">
</my-nameinput>
```

will be rendered as

``` html
<my-nameinput>
     Firstname: <input id="first">
     Lastname: <input id="last">
</my-nameinput>
```

But if the order in the `#shadow-root` is flipped, rendering will be

``` html
<my-nameinput>
     Lastname: <input id="last">
     Firstname: <input id="first">
</my-nameinput>
```

With the current spec, TAB navigation order is fixed for both, `#first` -> `#last`, as in the order of
the document DOM tree. But preferably, the order should be `#first` -> `#last` for the former and `#last` -> `#first` for the latter.

Is this the bottomline?
I feel we can define this without nested browsing context, just with composed tree.

[...]

PR #381, opened 2016-02-12
Proposed change for Focus Navigation

TakayoshiKochi:
As discussed in
https://github.com/w3c/webcomponents/issues/375

This adds a clarification about focus navigation order for
nodes distributed under a slot element.

Also added handling tabindex="-1" case.


TakayoshiKochi:
@hayatoito PTAL

spec/shadow/index.html
```diff
@@ -1139,6 +1141,27 @@
           </li>
         </ol>
 
+        <p>Likewise, the <a>sequential focus navigation order</a> for <a>distributed nodes</a>
```
hayatoito: Wrong... This should not be _distributed nodes_. This should be _assigned nodes_.

In addition to that, you should consider the followings:
1. Do not forget to include descendants of assigned nodes into this focus navigation group.
2. Consider the fallback contents of slots, in addition to assigned nodes.
3. When you try to consider 1 and 2, do not forget the case where a slot is a descendant of another slot.
4. The traversal order is not defined for _sequential focus navigation order for distributed nodes_. Define the traversal order of this focus navigation group.
5. Exclude an assigned node (and more) from its root node's focus navigation group. One node should not join more than one focus navigation group.
6. Mention how tabindex values are _scoped_ at each focus navigation order.

spec/shadow/index.html
```diff
@@ -1139,6 +1141,27 @@
...
+            <strong>must</strong> be inserted in place of <var>A</var> as if <var>A</var>
+            were assigned the <a>tabindex</a> value 0 for determining its position.</li>
+          <li>Otherwise the <a>sequential focus navigation order</a> for its
+            <a>distributed nodes</a> <strong>must</strong> be inserted in place of
+            <var>A</var> according to its <a>tabindex</a> value.</li>
+        </ol>
+        <p class="note">When a <a>slot</a> <var>A</var> is assigned to another
+          <a>slot</a> <var>B</var>,
+          the <a>sequential focus navigation order</a> of <var>A</var>'s <a>distributed nodes</a>
+          is inserted into <var>B</var>'s <a>sequential focus navigation order</a> following
```
hayatoito: We do not need to mention this.
Please try to have a general rule which covers this case.

spec/shadow/index.html
```diff
@@ -1139,6 +1141,27 @@
...
+            <strong>must</strong> be skipped if A is assigned <a>tabindex</a> value -1.
+          <li>If <var>A</var> is not given <a>tabindex</a>,
+            The <a>sequential focus navigation order</a> for its <a>distributed nodes</a>
+            <strong>must</strong> be inserted in place of <var>A</var> as if <var>A</var>
+            were assigned the <a>tabindex</a> value 0 for determining its position.</li>
+          <li>Otherwise the <a>sequential focus navigation order</a> for its
+            <a>distributed nodes</a> <strong>must</strong> be inserted in place of
+            <var>A</var> according to its <a>tabindex</a> value.</li>
+        </ol>
+        <p class="note">When a <a>slot</a> <var>A</var> is assigned to another
```
hayatoito: We do not need to mention this. Please try to have a _general_ rule which covers all cases.

Unless a good general rule, the spec would have a lot of cases which are not covered.
e.g.

``` html
<slot>
  <input></input>
  <input>
    <slot>
       <input><input>
     </slot>
  </input>
</slot>
```

spec/shadow/index.html
hayatoito: Wrong. A's scope is not flattened into B.

spec/shadow/index.html
```diff
@@ -1139,6 +1141,27 @@
...
+        <ol>
+          <li>The <a>sequential focus navigation order</a> for its <a>distributed nodes</a>
+            <strong>must</strong> be skipped if A is assigned <a>tabindex</a> value -1.
+          <li>If <var>A</var> is not given <a>tabindex</a>,
+            The <a>sequential focus navigation order</a> for its <a>distributed nodes</a>
+            <strong>must</strong> be inserted in place of <var>A</var> as if <var>A</var>
+            were assigned the <a>tabindex</a> value 0 for determining its position.</li>
+          <li>Otherwise the <a>sequential focus navigation order</a> for its
+            <a>distributed nodes</a> <strong>must</strong> be inserted in place of
+            <var>A</var> according to its <a>tabindex</a> value.</li>
```
hayatoito: You might want to clarify what "according to it's tabindex value" means.

A slot element is not focusable element. Thus, you should mention how tabindex value of a slot element is _effectively_ used in focus navigation order more clearly.

spec/shadow/index.html
TakayoshiKochi: Thanks.

Probably we want to start using '[control group](https://html.spec.whatwg.org/multipage/interaction.html#control-group)' (defined in HTML spec) for the scope.
I am working on rewriting with it.

TakayoshiKochi:
@hayatoito PTAL.

I tried to get this upstreamable to HTML spec, [section 6.4](https://html.spec.whatwg.org/multipage/interaction.html#focus),
but it is hard to do in a single step, as the HTML spec describes the
navigation order as an algorithm to iteratively find the next focusable node,
while the Shadow DOM spec (before my modification) assumes there is a
(somewhat) pre-computed _focus navigation order_ for each scope and
describes how they are merged together.

So this is my attempt to include the #375 as exactly as possible by
following the current way of describing focus navigation in Shadow DOM.

At this point it seems to me to patch the upstream spec requires lot more work.

spec/shadow/index.html
```diff
@@ -1123,13 +1123,46 @@
       </section>
 
       <section>
-        <h3>Focus Navigation</h3>
+        <h3>Sequential Focus Navigation</h3>
 
         <p>If a <a>node</a> doesn’t <a data-lt="participates">participate</a> in the <a>document flat tree</a>, the <a>node</a> <strong>must</strong> be skipped from the <a>sequential focus navigation</a>.</p>
 
-        <p>The <a>sequential focus navigation order</a> for a given <a>shadow tree</a> <var>A</var> <strong>must</strong> be inserted into the <a>sequential focus navigation order</a> for the <a>parent tree</a> <var>B</var> as follows:</p>
+        <p>Each <a>shadow root</a> and <a>slot</a> element is a <a>control group owner</a> and
```
hayatoito: A control group is more upper layer concept, isn't it?

I mean:
- There is 1 : N relationship between [a control group] and [focus navigation order in each scope].
- There is 1 : 1 relationship between [a control group] and [_global_ focus navigation order], which is the result of _merging_ multiple [focus navigation order in each scope].

spec/shadow/index.html
```diff
@@ -1123,13 +1123,46 @@
...
+          The order is tree order.
+        </p>
+
+        <p>For a given <a>slot</a> element,
+          its <a>sequential focus navigation order</a> consists of
+          <a data-lt="focusable area">focusable areas</a> of its <a>assigned nodes</a>
+          and descendants of the <a>assigned nodes</a>.
+          If a given slot's assigned nodes is empty, its
+          <a>sequential focus navigation order</a> consists of
+          <a data-lt="focusable area">focusable areas</a> of its fallback contents.
```
hayatoito: Fallback contents are not well defined.
Also do not forget that `<slot>` can be a descendant of another `<slot>`. The current definition does not cover this case at all.

hayatoito:
I think you will not get a good definition by the current approach. I recommend you to have a more formal approach:

e.g.

algorithm:
- Input: a document composed tree
- Output: set of owners, where each owner owns a group

algorithm:
- Input: node, NODE, in a document composed tree
- Output: a group which NODE belongs to

Thus, we can define the disjoint sets clearly. Then,

algorithm:
- Input: node A and node B where A nad B joins the same group
- Output: A or B, which comes before in the group

Thus, the _oder_ is clearly defined.

spec/shadow/index.html
TakayoshiKochi: My understanding is that frames own its own control group (e.g. `<iframe>`) and its own sequence,  but from user's standpoint, tab navigation sequence look _global_.  Thus sequences for each group is merged.

TakayoshiKochi:
It looks iterative reviews won't work well if completely different structure is suggested.

I've written the draft following your suggestion:
https://docs.google.com/document/d/1mNV2Z231SrAZuMJvBVNfVNHfULkMwUZOZ75vZM1K9zY/edit?usp=sharing (anyone can comment)

For the 3rd algorithm (ordering) in the comment,
for document/shadow root scope, the order is probably obvious (tree order),
but for slot scope (not fallback content scope) the traversal order is depth-first
search on result of getAssignedNodes()?

In addition, do we want to describe how to compare 2 focusable areas with tabindex here
to get the scoped order of the sequence?

That would add almost dead-copy of [HTML 6.4.3](https://html.spec.whatwg.org/multipage/interaction.html#the-tabindex-attribute) and
super redundant.

TakayoshiKochi:
@hayatoito comment on the doc linked above is welcome.

TakayoshiKochi:
@hayatoito
Updated the draft:
https://docs.google.com/document/d/1mNV2Z231SrAZuMJvBVNfVNHfULkMwUZOZ75vZM1K9zY/edit?usp=sharing (anyone can comment)

This should be better aligned with what you suggested... any feedback is welcome.
Sorry for the long turnaround.

TakayoshiKochi:
@hayatoito Updated the draft. PTAL.

TakayoshiKochi:
@hayatoito this branch went into some weird state, so I created another PR #447.
Could you take a look?

PR #396, opened 2016-02-24
Switched custom spec to respec

plehegar:
- Now with normative and informative references section
  -> removed dependencies section since it's no longer needed as-is
- Renamed section "About this Document" into "Conformance" to accommodate automatic triggers (such as RFC2119 text)
- Harmonized ES links to use tc39.github.io
- Use CSS3 selectors reference instead of CSS2
- s/pseudoclass/pseudo-class/
- Motivations section now appears first
- Use the linking to definitions of respec
- Use continuous IDL mode
- Switch to experimental style
- Added a manifest file for the automatic publishing system
- Various links added into the headers, eg link to Can I Use?
- Removed dependency to separate style sheet (copied 2 rules inline instead)
- sections Custom Tag Example and Type Extension Example are informative

### DIFF
--- a/spec/custom/W3CTRMANIFEST
+++ b/spec/custom/W3CTRMANIFEST
@@ -1,0 +1,2 @@
+index.html?specStatus=WD;shortName=custom-elements;useExperimentalStyles=false respec
+custom-elements-whole-world.svg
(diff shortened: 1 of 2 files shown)


domenic:
Looking at https://rawgit.com/w3c/webcomponents/custom/respec/spec/custom/index.html#normative-references the DOM, ES, and HTML Imports references are wrong.

plehegar:
I switched the ES and DOm ref but the HTML-IMPORTS will get fixed once html import is switched to automatic publishing (ongoing)

domenic:
:+1:

plehegar:
@dglazkov ok to merge?

dglazkov:
:+1:

Issue #365, opened 2016-01-25
Rename registerElement and change API slightly


domenic:
Per F2F: name it defineElement, and have it not return anything.

rniwa:
`defineElement` takes a tag name and an author defined constructor/class as arguments, and retains the author defined class for element construction.  Also see the issue #403 for the actual construction algorithm.

PR #400, opened 2016-02-26
s/document.registerElement/document.defineElement and no return value

plehegar:
Addresses #365
The spec still talks about element registration, registry of element definitions and has the type ElementRegistrationOptions but I think we can keep those.


domenic:
I don't think we should do this until we also implement the changed semantics (which are pretty pervasive) regarding element constructors and the HTMLElement constructor, discussed recently on public-webapps. Otherwise it gives a false impression that the consensus API is ready for implementation. It's better to keep the document in the old state until the new one is ready.

rniwa:
I agree with @domenic. We should probably rewrite the whole section.

plehegar:
ok. closing this one without merging then. Feel free to reopen if you believe otherwise.

Issue #165, opened 2015-07-06
[Custom]: Integrate with the HTML Parser


hayatoito:
Title: [Custom]: Make "Parsing Custom Elements" section more explicit (bugzilla: 27768)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27768

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27768#c0
_Anne_ wrote on 2015-01-06 16:13:07 +0000.

This section does not define how these elements are instantiated.

I would assume these to be added to the unresolved set of elements, no?

rniwa:
Per Jan F2F, parser constructs HTML elements synchronously. We need to rewrite this whole section.

Issue #239, opened 2015-07-06
[Custom]: "are" custom element names ASCII characters, or MUST they be ASCII characters?


hayatoito:
Title: [Custom]: "are" custom element names ASCII characters, or MUST they be ASCII characters? (bugzilla: 22056)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=22056

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=22056#c0
_Dominic Cooney_ wrote on 2013-05-16 06:29:42 +0000.

"is a sequence of alphanumeric ASCII characters"

This is confusing. NCName [1] includes combining characters and extenders that are not ASCII characters. These should be allowed, because custom element names MUST match the NCName production and there is no restriction on the character set.

I think "is a sequence of alphanumeric ASCII characters" should be "MUST be a sequence of ASCII characters".

[1] <http://www.w3.org/TR/1999/REC-xml-names-19990114/#NT-NCName\>

rniwa:
The current specification doesn't even mention ASCII characters anywhere:

https://w3c.github.io/webcomponents/spec/custom/#dfn-custom-element-type

> The custom element type identifies a custom element interface and is a sequence of characters that must match the NCName production [XML-NAMES], must contain a U+002D HYPHEN-MINUS character, and must not contain any uppercase ASCII letters [HTML].

I think it makes sense to restrict the tag name to ascii letters at least in v1.

rniwa:
Any opinions? @annevk @travisleithead @hober

notwaldorf:
I'd like to cast my vote for non-ascii in custom elements names! Emoji aside (https://jsbin.com/buzegi/edit?html,output) which is kind of cool, I think Kanji characters in tag names is a real use case 😊

Does the parser care if the tag name is non-ascii? Like, what makes it a hard problem (out of curiosity)?

rniwa:
See the issue #177. There are disagreements on the exact set of characters allowed in tag names.

Someone needs to investigate the issue and come up with a safe/correct subset of characters that can be used in custom elements.

I'm suggesting to restrict it to only ASCII characters in v1 since it's always safe to expand the set of characters being allowed latter once someone has done that work but not vice versa.

rniwa:
[...]

PR #405, opened 2016-03-01
Large custom element spec rewrite to implement some F2F decisions

### LINKED ISSUE #133: Remove the support for inherting from builtin subclasses of HTMLElement and SVGElement (bugzilla: 28547)
Title: Remove the support for inherting from builtin subclasses of HTMLElement and SVGElement (bugzilla: 28547)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28547

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28547#c0
_Ryosuke Niwa_ wrote on 2015-04-22 23:54:56 +0000.

See https://wiki.whatwg.org/wiki/Custom_Elements#Subclassing_existing_elements

Subclassing existing elements is hard as implementation-wise identity is both object-based and name / namespace based. Therefore subclassing an existing element (currently) requires that the name / namespace does not change. See also DOM: element constructors.

A hack was invented to make this work: <button is="my-button">. That hack is not well liked leaving us two options:

We leave this for now and work on this in parallel while stabilizing a smaller subset of custom elements.
We block on this and delay even more.
(Assuming that not all implementers are suddenly going to be okay with this hack.)

However, without this hack accessibility for trivial components is harder as more things have to be done by hand.

Another use case had emerged for the "is" hack: piggy-backing on parser behaviors. For example, extending <template> for data binding or as a way to specify shadow trees in HTML.

---

comment: 1
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28547#c1
_Martin Naumann_ wrote on 2015-04-23 10:16:51 +0000.

Please don't kill this. 

The fact that it's important to provide accessibility features is one aspect of why this is important and useful. 

The other aspect (and that's something that has actually bitten me in a live scenario) is defensiveness and progressive enhancement. 

I'd like to give an example of what happened to me, hurt my users and left me feel like an idiot for not using "is" extension. 

I wrote a component that wrapped Ace-Editor to provide syntax-highlighted code editing textareas. 

However, being not fully aware of the power behind the extension features, I built my component from scratch, using the textarea prototype, so

<code-editor></code-editor>

instead of 

<textarea is="code-editor"></textarea>

That worked nice, until I had a shaky mobile connection and the javascript didn't fully load. it left me with an unusable application. unfortunately it was not me finding this first, instead a bunch of people who use the internal tool were left with something that didn't work right when the connection was wonky or some other script had a parsing problem that prevented registration of the component. 

I quickly rectified the situation by correctly extending from a textarea, so the old fashioned Form submit action could be edited and POSTed, even when all the lovely enhancements that we are using (relying on Javascript) fail. 

Unless provided with a similar, internally less hacky, solution this should not be taken away from Web components. 

I think it's a fundamental trait of the web platform to be resilient and making Web components depend on javascript is harmful to this trait. 

Arguably we could leave it up to developers to provide those fullbacks outside of Web components, but that's making "doing the right thing" harder, which IMHO isn't a good idea either.

---

comment: 2
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28547#c2
_Ryosuke Niwa_ wrote on 2015-04-23 10:30:46 +0000.

The right fallback mechanism for web components is to use compostion as done in video, embed, object, and numerous other HTML elements.

e.g.
<code-editor><textarea></textarea></code-editor>

---

comment: 3
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28547#c3
_Martin Naumann_ wrote on 2015-04-23 10:43:25 +0000.

(In reply to Ryosuke Niwa from comment #2)

> The right fallback mechanism for web components is to use compostion as done
> in video, embed, object, and numerous other HTML elements.
> 
> e.g.
> <code-editor><textarea></textarea></code-editor>

Good point and definitely worth considering in this scenario, but it puts additional effort on the author using the component as they would have to always supply this nested element while the is-mechanism works without an additional level of content. At least that's what I feel when seeing this example.

Video, Embed, Object and Canvas provide alternative content in case one type of content isn't available.

The code-editor on the other hand enhances a feature that is a part of the platform, the textarea. I as an author prefer the extension mechanism over the composition.

Plus I am not entirely sure how that works with the accessibility affordances I would have on a browser that does support the component. would I still have to deal with ARIA and all of the things I'm getting "for free" with the extension?

---

comment: 4
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28547#c4
_Ryosuke Niwa_ wrote on 2015-05-14 07:39:40 +0000.

(In reply to Martin Naumann from comment #3)

> Plus I am not entirely sure how that works with the accessibility
> affordances I would have on a browser that does support the component. would
> I still have to deal with ARIA and all of the things I'm getting "for free"
> with the extension?

A lot of people are making this argument but that's simply not true. UAs can't make subclassed elements more accessible automatically.

### LINKED ISSUE #134: Custom elements should not upgrade elements by setting prototype (bugzilla: 28544)
Title: Custom elements should not upgrade elements by setting prototype (bugzilla: 28544)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28544

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28544#c0
_Ryosuke Niwa_ wrote on 2015-04-22 23:45:08 +0000.

Discussed here:
https://lists.w3.org/Archives/Public/public-webapps/2014JanMar/0158.html
https://lists.w3.org/Archives/Public/public-webapps/2015JanMar/0083.html

We shouldn't support upgrading of existing elements since it leaves the corresponding DOM objects with a wrong identity.

This is also not a normal programming model. Even with ES6 modules, we don't create a dummy modules while modules are loading and later automatically replace them under the foot.

---

comment: 1
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28544#c1
_Domenic Denicola_ wrote on 2015-04-22 23:52:35 +0000.

> Even with ES6 modules, we don't create a dummy modules while modules are loading and later automatically replace them under the foot.

This is exactly what we do with ES6 modules.

---

comment: 2
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28544#c2
_Ryosuke Niwa_ wrote on 2015-04-22 23:57:31 +0000.

(In reply to Domenic Denicola from comment #1)

> > Even with ES6 modules, we don't create a dummy modules while modules are loading and later automatically replace them under the foot.
> 
> This is exactly what we do with ES6 modules.

How? import { myClass } from 'mymodule'; doesn't create a promise or anything. It'll just block until mymodule is loaded.

---

comment: 3
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28544#c3
_Domenic Denicola_ wrote on 2015-04-22 23:58:58 +0000.

Referencing myClass will be a reference error until 'mymodule' is loaded. This normally comes up in the case of circular dependencies.

---

comment: 4
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28544#c4
_Ryosuke Niwa_ wrote on 2015-04-23 00:35:18 +0000.

(In reply to Domenic Denicola from comment #3)

> Referencing myClass will be a reference error until 'mymodule' is loaded.
> This normally comes up in the case of circular dependencies.

I'm not certain what you mean by "until 'mymodule' is loaded' but the 'mymodule' will be synchronously fetched and evaluated before the following statements are executed.

e.g.
import { myClass } from 'mymodule'; // (1)
new myClass; // this line of code is never executed until (1) is done executing.

The fact we can encounter a circular dependency in (1) is an orthogonal issue. The fact of matter is that the declarative syntax for importing an ES6 module DOES synchronously block the script execution.

---

comment: 5
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28544#c5
_Domenic Denicola_ wrote on 2015-04-23 01:23:23 +0000.

It is correct that (2) will not execute until (1) is done executing. But (1) simply creates a dummy myClass binding which will cause ReferenceErrors if executed before myClass's definition is evaluated. myClass's definition may or may not be evaluated by executing (1).

Here is a simple example:

// a.js
import \* as bModule from './b.js';
import { b } from './b.js';

// bModule is a dummy module with no properties
// if b is referenced it will cause a ReferenceError
// console.log(b); // fails

export let a = 1;

setTimeout(() => {
  // bModule is no longer a dummy module and has a b property
  // b can be referenced without errors
  console.log(b); // works
}, 100);

////////

// b.js
import \* as aModule from './a.js';
import { a } from './a.js';

export let b = 2;

////////

// entry.js
import './b.js';

Execution order is entry.js -> a.js -> b.js. Thus inside a.js we have the dummy module which later gets upgraded. In fact this happens for a.js (and entry.js) as well, but you don't observe that since you can't see them pre-upgrade.

---

comment: 6
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28544#c6
_Ryosuke Niwa_ wrote on 2015-04-23 02:09:24 +0000.

Again, even in this example, files are fetched and evaluated synchronously.

But let us not derail this bug from the main topic, which is about not upgrading custom elements. Even if my statement about class wasn't accurate, the fact that upgrading existing elements is a bad idea wouldn't change.

---

comment: 7
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28544#c7
_Domenic Denicola_ wrote on 2015-04-23 02:11:11 +0000.

I think it's very relevant. In a dynamic system with interdependencies, upgrading of some sort is important. We have it for ES6 modules, and it makes just as much sense for custom elements, for the same reasons.

---

comment: 8
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28544#c8
_Ryosuke Niwa_ wrote on 2015-04-23 02:42:25 +0000.

(In reply to Domenic Denicola from comment #7)

> I think it's very relevant. In a dynamic system with interdependencies,
> upgrading of some sort is important. We have it for ES6 modules, and it
> makes just as much sense for custom elements, for the same reasons.

Why?

### LINKED ISSUE #137: Custom elements should call user defined constructor synchronously (bugzilla: 28543)
Title: Custom elements should call user defined constructor synchronously (bugzilla: 28543)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28543

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=28543#c0
_Ryosuke Niwa_ wrote on 2015-04-22 23:38:23 +0000.

See https://lists.w3.org/Archives/Public/public-webapps/2015JanMar/0731.html

We should make the custom element's constructor be called synchrnously instead of at the end of the micro-task since that's not consistent with builtin elements and exposes "unconstructed" elements in some cases; e.g. immediately after cloning or setting innerHTML.

### LINKED ISSUE #158: [Custom]: "All Algorithms in One Diagram" (bugzilla: 27780)
Title: [Custom]: "All Algorithms in One Diagram" (bugzilla: 27780)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27780

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27780#c0
_Anne_ wrote on 2015-01-07 15:06:15 +0000.

Per this diagram the prototype is changed after any callbacks are invoked but that is wrong per the prose.

---

comment: 1
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27780#c1
_Dimitri Glazkov_ wrote on 2015-01-08 19:19:14 +0000.

Is this diagram useful, btw? I can't decide. My hesitation to remove it may be colored by hours I spent hand-coding it :P

---

comment: 2
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27780#c2
_Anne_ wrote on 2015-01-09 06:57:26 +0000.

I would love more diagrams in specifications and once I find a sensible way to generate them I would include them elsewhere too. It helps a lot with checking the overall flow or interdependencies.

---

comment: 3
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27780#c3
_Michael[tm] Smith_ wrote on 2015-01-09 20:25:41 +0000.

(In reply to Dimitri Glazkov from comment #1)

> Is this diagram useful, btw? I can't decide. My hesitation to remove it may
> be colored by hours I spent hand-coding it :P

(In reply to Anne from comment #2)

> I would love more diagrams in specifications

Agreed with Anne here. I think in general spec readers/reviewers/implementors really appreciate and benefit from good diagrams, so trying to provide good ones is a great goal.

(And I think diagrams are one of those things that users of the spec often end up just taking for granted when they have them; it's one of those many things that reviewers don't take time to make positive comments about. Instead, you're only going to hear from them when there's something wrong in the diagram. That suggests that if you want to measure whether a diagram is being used, you should put some easter-eggish error into it to see how many people notice and comment...)

### LINKED ISSUE #161: [Custom]: Improve the wording in element upgrade algorithm (bugzilla: 27770)
Title: [Custom]: Improve the wording in element upgrade algorithm (bugzilla: 27770)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27770

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27770#c0
_Dimitri Glazkov_ wrote on 2015-01-06 17:24:30 +0000.

It's not super-clear that the algorithm actually updates the candidates map.

### LINKED ISSUE #162: [Custom]: Make ES6 prose normative when ES6 ships (bugzilla: 25669)
Title: [Custom]: Make ES6 prose normative when ES6 ships (bugzilla: 25669)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=25669

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=25669#c0
_Dimitri Glazkov_ wrote on 2014-05-12 22:23:34 +0000.

Once ES6 is released, we need to turn the informative section at:

http://w3c.github.io/webcomponents/spec/custom/#es6

into normative prose.

### LINKED ISSUE #163: [Custom]: Convert all ES5 references to ES6 (bugzilla: 27405)
Title: [Custom]: Convert all ES5 references to ES6 (bugzilla: 27405)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27405

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27405#c0
_Dimitri Glazkov_ wrote on 2014-11-21 22:23:45 +0000.

See bug 26365.

### LINKED ISSUE #167: [Custom]: Specify preference of localName and typeExtension in createElement (bugzilla: 26852)
Title: [Custom]: Specify preference of localName and typeExtension in createElement (bugzilla: 26852)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=26852

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=26852#c0
_Deepak Singla_ wrote on 2014-09-19 10:00:15 +0000.

As specified in instantiation of custom elements that "custom tag" will take preference over "type extension", it is required to specify the same in case of createElement(...) .

 Element createElement(DOMString localName, DOMString typeExtension);

Here if localName is also a custom tag, then what will be the prototype of "Element"? Will it be of typeExtension or of localName.

Consider the following case:

var xA = doc.registerElement("x-a");
var xB = doc.registerElement("x-b");

var customElement = doc.createElement("x-a", "x-b");

What should be the prototype of customElement?

AFAIK, this will be the same case as specifying:

<x-a is="x-b"></x-a>

In this case prototype will be of "x-a".

If the above theory is correct, then web-platform-tests/custom-elements/instantiating-custom-elements/extensions-to-document-interface/create-element-interface-type-is-a-type-extension.html test case is wrong.

If this issue has been raised earlier, then please provide the link and close this bug.

### LINKED ISSUE #169: [Custom]: Section 10 prototype is not defined (bugzilla: 27643)
Title: [Custom]: Section 10 prototype is not defined (bugzilla: 27643)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27643

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27643#c0
_Erik Arvidsson_ wrote on 2014-12-17 22:07:55 +0000.

Let ERROR and DEFINITION be the result of running definition construction algorithm with DOCUMENT, TYPE, PROTOTYPE, and NAME as arguments

There is no PROTOTYPE here.

### LINKED ISSUE #170: [Custom]: SVG diagram accessibility (bugzilla: 27611)
Title: [Custom]: SVG diagram accessibility (bugzilla: 27611)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27611

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27611#c0
_steve faulkner_ wrote on 2014-12-15 12:17:10 +0000.

"Appendix A: All Algorithms in One Diagram"
the svg Diagram is not accessible, will look into making it so.

---

comment: 1
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27611#c1
_steve faulkner_ wrote on 2014-12-15 12:21:37 +0000.

http://w3c.github.io/webcomponents/spec/custom/#appendix-a

### LINKED ISSUE #185: [Custom]: Need to define global/Realm handling for registerElement (bugzilla: 27260)
Title: [Custom]: Need to define global/Realm handling for registerElement (bugzilla: 27260)

Migrated from: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27260

---

comment: 0
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27260#c0
_Boris Zbarsky_ wrote on 2014-11-06 17:18:44 +0000.

Created attachment 1535
Testcase

Consider this testcase:

  <iframe></iframe>
  <script>
    var f = document.registerElement.call(frames[0].document, "x-foo");
  </script>

Per the spec, this should "let PROTOTYPE be the result of invoking Object.create with HTMLElement's interface prototype object as only argument".  But it doesn't say _which_ "HTMLElement's interface prototype object" should be used.

Then the spec says to run the "custom element constructor generation algorithm", but this algorithm assumes an ambient global, which the spec doesn't define.

In practice, the Chrome and Firefox implementations differ, as the attached testcase shows.  The output in Chrome:

  ctor instanceof iframe Function: false
  ctor instanceof parent Function: true
  proto comes from iframe: false
  proto comes from parent: true

The output in Firefox:

  ctor instanceof iframe Function: true
  ctor instanceof parent Function: false
  proto comes from iframe: true
  proto comes from parent: false

I believe the Firefox behavior is more correct here, by the way, since we're supposed to be using the registry of the iframe document, so should use its prototypes and constructors.

---

comment: 1
comment_url: https://www.w3.org/Bugs/Public/show_bug.cgi?id=27260#c1
_Boris Zbarsky_ wrote on 2014-11-06 17:26:09 +0000.

Oh, and also, the "definition construction algorithm" is making assumptions about which global's prototypes or interfaces are supposed to be used.  This part is also not interoperable in Chrome and Firefox: Firefox seems to require that the passed-in proto come from the same global as the document involved, as you can see in the testcase.

### LINKED ISSUE #283: #dfn-set-prototype is unused
I cannot find where objects' [[Prototype]]s are set during upgrading.

### LINKED ISSUE #365: Rename registerElement and change API slightly

### LINKED ISSUE #403: Rewrite Custom Element Construction Algorithm
Here are steps to construct a custom element as agreed during Jan F2F as I promised to write down [1] [2]:

Modify http://w3c.github.io/webcomponents/spec/custom/#dfn-element-definition as follows:
The element definition describes a custom element and consists of:
- custom element type,
- local name,
- namespace,
- custom element interface,
- lifecycle callbacks, and
- element construction stack.

Each element construction stack is initially empty, and each entry is an instance of Element or a "AlreadyConstructed" marker.

Non-Normative Note: We need a stack per element definition to allow construction of other custom elements inside a custom element's constructor. Without such a stack per element definition, we would end up walking through entries in the stack to find the "right" entry.  Implementors are free to take such an approach to minimize the memory usage, etc..., but there are a lot of edge cases that need to be taken care of, and it's not a great way to spec an interoperable behavior.
## Custom Element Construction Algorithm

**Input**
- NAME, the custom element name.
- DOCUMENT, the owner document for new custom element.
- EXOTIC-TARGET, the target Element to be constructed / upgraded.

**Output**
- ELEMENT, new custom element instance.
- ERROR, could be either "None", "NotFound", "InvalidStateError", or an ECMAScript exception.
1. Let ERROR be "None".
2. Let REGISTRY be the (custom element) registry of DOCUMENT.
3. If DOCUMENT is an HTML document, let NAME be converted to ASCII lowercase.
4. Let DEFINITION be the element definition of with the local name, NAME, in REGISTRY.
5. If there is no matching definition, set ERROR to "NotFound" and terminate these steps.
6. Otherwise, push a new entry, EXOTIC-TARGET, to the element construction stack of DEFINITION.
7. Invoke the [[Construct]] internal method [3] on custom element interface of DEFINITION.
8. Pop the entry from the element construction stack of DEFINITION.
9. If the [[Construct]] invocation resulted in an exception, set ERROR to the raised exception, and terminate these steps.
10. Otherwise, let ELEMENT be the result of the invocation.
11. If ELEMENT is not the same Object value as EXOTIC-TARGET, set ERROR to "InvalidStateError", and terminate these steps.

_Non-Normative Note_: we can modify step 4 to support non-HTML elements in the future. In step 11, ELEMENT can be different from EXOTIC-TARGET if the custom element's constructor instantiates another instance of the same custom element before calling super().
## HTMLElement constructor

Non-Normative Note: HTMLElement's constructor is called via super() call inside the custom element constructor.
1. Let TARGET be GetNewTarget(). [4]
2. Let DOCUMENT be the associated document of the global object (the result of GetGlobalObject() [5]).
3. Let REGISTRY be the (custom element) registry of DOCUMENT.
4. Let DEFINITION be the element definition with the element interface, TARGET, in REGISTRY.
5. If there is no matching definition, throw TypeError and terminate these steps.
6. Let NAME be the local name of DEFINITION.
7. If the element construction stack of DEFINITION is empty,
   1. Return a new element that implements HTMLElement, with no attributes, namespace set to the HTML namespace,
      local name set to NAME, and node document set to DOCUMENT.
8. Otherwise, let INSTANCE be the last entry in the element construction stack (i.e. in LIFO).
9. If INSTANCE is a "AlreadyConstructed" marker, throw InvalidStateError and terminate these steps.
10. Otherwise, replace the last entry in the element construction stack with a "AlreadyConstructed" marker.
11. Return INSTANCE.

Non-Normative Note: step 7.1. is like step 4 in createElement [6] and happens when author script instantiates a custom element without going through DOM. e.g. "new X". Checks in Step 9 and 10 are needed when author scripts constructs invokes super() multiple times inside a custom element constructor. Step 9 is sufficient for the Custom Element Construction Algorithm to fail because it checks the exception in step 9. Step 5 could throw NotSupportedError instead if people would prefer that.

[1] https://github.com/w3c/WebPlatformWG/blob/gh-pages/meetings/25janWC.md
[2] https://www.w3.org/2016/01/25-webapps-minutes.html
[3] http://www.ecma-international.org/ecma-262/6.0/#sec-ecmascript-function-objects-construct-argumentslist-newtarget
[4] http://www.ecma-international.org/ecma-262/6.0/#sec-getnewtarget
[5] http://www.ecma-international.org/ecma-262/6.0/#sec-getglobalobject
[6] https://dom.spec.whatwg.org/#dom-document-createelement

domenic:
So, I accidentally rewrote a lot of the spec. All I wanted to do was land @rniwa's algorithm from #403 in, but it turns out everything is connected to everything else, and some sections were just so tempting, and... well, a lot of changes happened ^_^.

View the end result at https://rawgit.com/w3c/webcomponents/construction/spec/custom/index.html

Not sure how much editorial review we want to do. This is a big improvement in terms of aligning with the F2F's results and moving toward clean patches onto DOM and HTML that we can move into those specs. It might be better to land it soon instead of making sure all the editorial stuff is perfect.

It would be good to get the concepts reviewed of course. @rniwa, @annevk, @travisleithead?

---

Notable changes:
- Implemented HTMLElement constructor using @rniwa's algorithm from #403.
- Rewrote element upgrading to use @rniwa's algorithm from #403, and incorporated it into the rest of the upgrading considerations.
- Got rid of the ability to extend SVGElement, thus allowing us to remove most mentions of namespaces from the spec.
- Removed createdCallback.
- Rewrote "registering elements":
  - Uses defineElement instead of registerElement
  - Preserves the constructor instead of grabbing the .prototype property and synthesizing a new constructor
  - No longer spread out over four separate algorithms plus registerElement; everything is now inline under defineElement
  - More precise about exactly how to get the custom element prototype and callbacks
- Rewrote createElement and createElementNS to be replacements instead of patches, and to call the author-supplied constructor.
- Removed the "All Algorithms in One Diagram" section since so many algorithms changed or were inlined into their callers.
- Removed the "Custom Elements and ECMAScript 6" section since it is very obsolete and does not reflect our current thinking.
- New and rewritten algorithms do not use the unorthodox INPUTS/OUTPUTS blocks, or capitalized variable names. This is kind of a nice marker of new vs. old content.

Notable things _not_ substantially changed:
- Parser changes are not specced still, besides to say that they should construct the element using its constructor.
- Lifecycle callbacks were not changed, except for removing createdCallback.
- Type extensions were not removed (yet?); it seems better to have a modernized version of them that we atomically remove.
- Registries were not made available everywhere.

Closes #403. Closes #365. Closes #283. Closes #185. Closes #170. Closes #169. Closes #167. Closes #163. Closes #162. Closes #161. Closes #158. Closes #137 (modulo the fact that #165 is still open). Closes #134. Closes #133.


travisleithead:
This is a great change. I'd love to see the few example scenarios moved up in the doc (so that the reader encounters them for context earlier on before getting totally lost in the algorithms). +1 to merging in and then continuing to work out editorial issues.

spec/custom/index.html
```diff
@@ -130,23 +130,21 @@
...
+<p>A <dfn id="dfn-element-definition">custom element definition</dfn> describes a <a>custom element</a> and consists of:</p>
 <ul>
-    <li><a>custom element type</a>,</li>
-    <li><a href="https://dom.spec.whatwg.org/#concept-element-local-name">local name</a> [[!WHATWG-DOM]],</li>
-    <li><a href="https://dom.spec.whatwg.org/#concept-element-namespace">namespace</a> [[!WHATWG-DOM]],</li>
-    <li><a>custom element prototype</a>, and </li>
-    <li><a>lifecycle callbacks</a>.</li>
+    <li>a <dfn id="dfn-element-definition-type" for="custom element definition">type</dfn>, which is a <a>custom element type</a>,</li>
+    <li>a <dfn id="dfn-element-definition-local-name" for="custom element definition">local name</dfn>, which is a <a href="https://dom.spec.whatwg.org/#concept-element-local-name">local name</a>,</li>
+    <li>a <dfn id="dfn-element-definition-constructor" for="custom element definition">konstructor</dfn>, which is a <a>custom element constructor</a>,</li> <!-- https://github.com/w3c/respec/issues/582 for why it's k -->
```
rniwa, line 136: Typo: konstructor.

travisleithead, line 136: Haha! I thought that was deliberate!

spec/custom/index.html
```diff
@@ -130,23 +130,21 @@
...
+<p>Each <a>custom element definition</a>'s <a href="#dfn-element-definition-construction-stack">construction stack</a> is manipulated by the <a>upgrade a custom element</a> algorithm and the <a href="#dom-htmlelement-constructor"><code>HTMLElement</code> constructor</a>. Each entry in the list will be either an <a href="https://dom.spec.whatwg.org/#concept-element">element</a> or an <dfn id="dfn-already-constructed-marker"><i>already constructed</i></dfn> marker.</p>
 
-<p><dfn id="dfn-element-registration">Element registration</dfn> is a process of adding an <a>element definition</a> to a <a>registry</a>. One <a>element definition</a> can only be <a data-lt="element registration">registered</a> with one <a>registry</a>.</p>
+<p>At the time of creation, a <a href="https://dom.spec.whatwg.org/#concept-document">document</a> could be associated with a <a>registry</a>. A <dfn id="dfn-registry">registry</dfn> is a <a href="https://en.wikipedia.org/wiki/Set_(computer_science)">set</a> of <a data-lt="custom element definition">custom element definitions</a>.</p>
```
rniwa, line 144: Could be associated?  Shouldn't this be always the case?

spec/custom/index.html
```diff
@@ -369,202 +326,118 @@
...
-    <li>Let <var>NAME</var> be <var>EXTENDS</var></li>
-    <li>Let <var>ERROR</var> be the result of running the <a>element registration algorithm</a> with <var>DOCUMENT</var>, <var>TYPE</var>, <var>PROTOTYPE</var>, and <var>NAME</var> as arguments</li>
-    <li>If <var>ERROR</var> is <code>InvalidType</code>, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>SyntaxError</code> and <strong>stop</strong>.</li>
-    <li>If <var>ERROR</var> is not <code>None</code>, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>NotSupportedError</code> and <strong>stop</strong>.</li>
-    <li>Return result of running <a data-lt="custom element constructor generation algorithm">custom element constructor generation algorithm</a> with <var>DOCUMENT</var> and <var>PROTOTYPE</var> as arguments.</li>
+<ol>
+    <li>Let <var>type</var> be <a href="https://dom.spec.whatwg.org/#converted-to-ascii-lowercase">converted to ASCII lowercase</a>.</li>
```
rniwa: This should only happen in a HTML document, not a XML document.
Also, aren't we disallowing/throwing when there is an uppercase ASCII letter in the tag name?
Also, I'd prefer calling this as tag name since that's what it is.

spec/custom/index.html
```diff
@@ -369,202 +326,118 @@
...
-<h3>Extensions to <a href="https://dom.spec.whatwg.org/#document"><code>Document</code></a> Interface</h3>
+<h3>New <a href="https://dom.spec.whatwg.org/#document"><code>Document</code></a> Method</h3>
 
-<p>The <dfn for='Document'>registerElement</dfn> method of the <a href="https://dom.spec.whatwg.org/#document">Document</a> interface provides a way to <a data-lt="element registration">register</a> a <a>custom element</a> and returns its <a>custom element constructor</a>.</p>
+<p>The <dfn for="Document">defineElement</dfn> method of the <a href="https://dom.spec.whatwg.org/#document">Document</a> interface provides a way to <a data-lt="element registration">register</a> a <a>custom element</a>.</p>
 
 <pre class='idl'>
 partial interface Document {
-    Function registerElement(DOMString type, optional ElementRegistrationOptions options);
+    Function defineElement(DOMString type, Function constructor, optional ElementRegistrationOptions options);
```
rniwa: defineElement should have "void" as the return type.

spec/custom/index.html
domenic, line 136: It is deliberate; see the HTML comment at the end of the line :( https://github.com/w3c/respec/issues/582

spec/custom/index.html
```diff
@@ -369,202 +326,118 @@
...
+        <ol>
+            <li>If the <a href="https://dom.spec.whatwg.org/#context-object">context object</a> is an <a href="https://dom.spec.whatwg.org/#html-document">HTML document</a>, let <var>extends</var> be <a href="https://dom.spec.whatwg.org/#converted-to-ascii-lowercase">converted to ASCII lowercase</a>.</li>
+
+            <li>If there is no <a href="https://dom.spec.whatwg.org/#concept-element-interface">element interface</a> for <var>extends</var> and the <a href="https://dom.spec.whatwg.org/#html-namespace">HTML namespace</a>, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>NotSupportedError</code> and abort these steps.</li>
+
+            <li>Set <var>localName</var> to <var>extends</var>.</li>
+        </ol>
+    </li>
+
+    <li>Let <var>prototype</var> be <a href="https://tc39.github.io/ecma262/#sec-get-o-p">Get</a>(<var>constructor</var>, "prototype"). Rethrow any exceptions.</li>
```
rniwa, line 386: We should also check that &lt;var>constructor&lt;/var> is indeed a constructor.

spec/custom/index.html
domenic, line 144: Probably. I didn't touch this much. I'm not sure why the diff viewer got confused but that just moved line 142 to line 145.

spec/custom/index.html
domenic: Good catch, thanks.

spec/custom/index.html
domenic: The latter two sound like good follow-up changes. Will make the first one here.

spec/custom/index.html
domenic, line 386: Great idea, thanks

rniwa, line 386: I also think it's valuable to check that the [[Prototype]] internal slot of `constructor` is HTMLElement (or inherits from HTMLElement when `extends` is specified) as an early error although the author can modify __proto__ after the fact so it might be sufficient to check when the element gets constructed.

Anyway, this change can probably be made separately.

spec/custom/index.html
```diff
@@ -584,167 +457,224 @@
...
+The <a href="https://html.spec.whatwg.org/multipage/dom.html#htmlelement"><code>HTMLElement</code></a> interface gains following annotation:
+
+<pre>
+[Constructor]
+</pre>
+
+The <dfn id="dom-htmlelement-constructor"><code>HTMLElement</code> constructor</dfn>, when invoked, must perform the following steps:
+
+<ol>
+    <li>Let <var>document</var> be the <a href="https://html.spec.whatwg.org/multipage/#concept-document-window">associated document</a> of <a href="https://html.spec.whatwg.org/#the-current-realm-record">the current Realm Record</a>'s [[\GlobalObject]].</li>
```
rniwa: As we discussed on IRC, I don't think this is quite right in the case a custom element extends another document's prototype (e.g. to have its own registry).  We should be using the global object of `[[Prototype]]` internal slot of the constructor or the global object of the realm of the `HTMLElement` constructor itself.

spec/custom/index.html
```diff
@@ -584,167 +457,224 @@
...
+
+    <li>
+        <p>Otherwise, if <var>registry</var> contains an entry with <a href="#dfn-element-definition-local-name">local name</a> <var>localName</var>:</p>
+
+        <ol>
+            <li>Let <var>definition</var> be the entry in <var>registry</var> with <a href="#dfn-element-definition-local-name">local name</a> <var>localName</var>.</li>
+
+            <li>Let <var>C</var> be <var>definition</var>'s <a href="#dfn-element-definition-constructor">constructor</a>.</li>
+
+            <li>Return the result of <a href="https://tc39.github.io/ecma262/#sec-construct">Construct</a>(<var>C</var>), rethrowing any exceptions.</li>
```
domenic, line 566: When speccing this I ended up just doing Construct(C) in createElement. I vaguely recall @annevk wanting to use the upgrading path for this, since it's a DOM method and so it should use the same nanotask infrastructure? @rniwa points out the difference is observable in edge cases where you construct yourself inside your constructor before calling super().

spec/custom/index.html
```diff
@@ -584,167 +459,225 @@
...
```
rniwa, line 566: We should check the return value of `Construct` call and throw or return `null` / `undefined` when it's not an `Element`.

spec/custom/index.html
```diff
@@ -584,167 +459,225 @@
...
+    <li>Let <var>prototype</var> be <var>definition</var>'s <a href="#dfn-element-definition-prototype">prototype</a>.</li>
 
-    <li>Let <var>NAME</var> be <var>EXTENDS</var></li>
-    <li>Let <var>ERROR</var> be the result of running the <a>element registration algorithm</a> with <var>DOCUMENT</var>, <var>TYPE</var>, <var>PROTOTYPE</var>, and <var>NAME</var> as arguments</li>
-    <li>If <var>ERROR</var> is <code>InvalidType</code>, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>SyntaxError</code> and <strong>stop</strong>.</li>
-    <li>If <var>ERROR</var> is not <code>None</code>, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>NotSupportedError</code> and <strong>stop</strong>.</li>
-    <li>Return result of running <a data-lt="custom element constructor generation algorithm">custom element constructor generation algorithm</a> with <var>PROTOTYPE</var>, <var>FUNCTION</var>, and <var>DOCUMENT</var> as arguments.</li>
-</ol>
-</div>
+    <li>Perform <var>element</var>.[[\SetPrototypeOf]](<var>prototype</var>). Rethrow any exceptions.</li>
```
rniwa: It's probably best if we performed this inside the HTMLElement constructor so that in the case the constructor throws before finishing super call, we wouldn't leave the element in a weird state.

It would also be more consistent with the way ES6 classes are constructed (or the way you'd naturally write polyfill).

spec/custom/index.html
```diff
@@ -584,167 +459,225 @@
...
 
-<p>Similarly, the <a data-lt="custom element constructor generation algorithm">custom element constructor generation algorithm</a> will change as follows:</p>
+    <li>Add <var>element</var> to the end of <var>definition</var>'s <a href="#dfn-element-definition-construction-stack">construction stack</a>.</li>
```
rniwa, line 673: Should we use terminologies like push and pop instead since it's a stack?

spec/custom/index.html
```diff
@@ -584,167 +459,225 @@
...
+    <li>Let <var>constructResult</var> be <a href="https://tc39.github.io/ecma262/#sec-construct">Construct</a>(<var>C</var>).</li>
+
+    <li>Remove <var>element</var> from the end of <var>definition</var>'s <a href="#dfn-element-definition-construction-stack">construction stack</a>.</li>
+
+    <li>If <var>constructResult</var> is an abrupt completion, return <var>constructResult</var> (i.e., re-throw the exception).</li>
+
+    <li>
+        <p>If <a href="https://tc39.github.io/ecma262/#sec-samevalue">SameValue</a>(<var>constructResult</var>.[[\value]], <var>element</var>) is false, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>InvalidStateError</code> and terminate these steps.</p>
+
+        <p class="note">This can occur if <var>C</var> constructs another instance of the same custom element before calling <code>super()</code>.</p>
```
rniwa: Or if the constructor returns a random object e.g. {}.

spec/custom/index.html
```diff
@@ -369,202 +326,120 @@
...
+
+    <li>If the <a href="https://dom.spec.whatwg.org/#context-object">context object</a> is an <a href="https://dom.spec.whatwg.org/#html-document">HTML document</a>, let <var>type</var> be <a href="https://dom.spec.whatwg.org/#converted-to-ascii-lowercase">converted to ASCII lowercase</a>.</li>
+
+    <li>If <var>type</var> is an invalid <a>custom element type</a>, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>SyntaxError</code> and abort these steps.</li>
+
+    <li>Let <var>registry</var> be the <a href="https://dom.spec.whatwg.org/#context-object">context object</a>'s <a>registry</a>. If <var>registry</var> does not exist, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>NotSupportedError</code> and abort these steps.</li>
+
+    <li>If <var>registry</var> contains an entry with <a href="#dfn-element-definition-type">custom element type</a> <var>type</var>, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>NotSupportedError</code> and abort these steps.</li>
```
rniwa: We also need to check any entry with `constructor` since it seems that we don't want to allow a single interface to be associated with multiple tag names at least in v1.

rniwa:
Great change overall!  Hopefully someone from Mozilla can glance over this and merge it tonight or tomorrow.

spec/custom/index.html
annevk, line 136: Seems easy to fix when upstreaming.

spec/custom/index.html
annevk, line 566: I don't think I had that concern. I'm happy to use synchronous where we can get away with it.

spec/custom/index.html
annevk, line 673: Add and remove is more consistent with ECMAScript.

spec/custom/index.html
```diff
@@ -112,7 +112,7 @@
 <section>
 <h2>Concepts</h2>
 
-<p><dfn id="dfn-custom-element">Custom element</dfn> is <a href="https://heycam.github.io/webidl/#dfn-platform-object">platform object</a> whose <a href="https://heycam.github.io/webidl/#dfn-interface">interface</a>  [[!WebIDL]] is defined by the author. The <a href="https://heycam.github.io/webidl/#interface-prototype-object">interface prototype object</a> of a <a>custom element</a>'s <a href="https://heycam.github.io/webidl/#dfn-interface">interface</a> is called the <dfn id="dfn-custom-element-prototype">custom element prototype</dfn>.</p>
+<p>A <dfn id="dfn-custom-element">custom element</dfn> is <a href="https://heycam.github.io/webidl/#dfn-platform-object">platform object</a> whose constructor and prototype is defined by the author. The constructor function supplied by the author is called the <dfn id="dfn-custom-element-constructor">custom element constructor</dfn>.</p>
```
annevk: a platform object*

spec/custom/index.html
```diff
@@ -80,7 +80,7 @@
 
 <section id='abstract'>
 
-<p>This specification describes the method for enabling the author to define and use new types of DOM elements in a document.</p>
+<p>This specification describes the method for enabling the author to define and use new types of DOM elements in a document. It will eventually be upstreamed into [[!WHATWG-DOM]] and [[!HTML]].</p>
```
annevk: And IDL.

spec/custom/index.html
```diff
@@ -369,202 +326,120 @@
...
 };
 
 dictionary ElementRegistrationOptions {
-     object? prototype = null;
      DOMString? extends = null;
 };
 </pre>
 
-<p>When called, the <a for='Document'>registerElement</a> method MUST run these steps:</p>
+<p>The <a for="Document">defineElement</a> method, when invoked, must run these steps:</p>
```
annevk: Editorial: include the method signature here for easy of reading.

spec/custom/index.html
```diff
@@ -369,202 +326,120 @@
...
+
+    <li>If the <a href="https://dom.spec.whatwg.org/#context-object">context object</a> is an <a href="https://dom.spec.whatwg.org/#html-document">HTML document</a>, let <var>type</var> be <a href="https://dom.spec.whatwg.org/#converted-to-ascii-lowercase">converted to ASCII lowercase</a>.</li>
+
+    <li>If <var>type</var> is an invalid <a>custom element type</a>, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>SyntaxError</code> and abort these steps.</li>
+
+    <li>Let <var>registry</var> be the <a href="https://dom.spec.whatwg.org/#context-object">context object</a>'s <a>registry</a>. If <var>registry</var> does not exist, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>NotSupportedError</code> and abort these steps.</li>
+
+    <li>If <var>registry</var> contains an entry with <a href="#dfn-element-definition-type">custom element type</a> <var>type</var>, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> a <code>NotSupportedError</code> and abort these steps.</li>
+
+    <li>Let <var>localName</var> be <var>type</var>.</li>
```
annevk: Editorial: why not localName from the start?

spec/custom/index.html
```diff
@@ -369,202 +326,120 @@
...
+
+    <li>Let <var>attributeChangedCallback</var> be <a href="https://tc39.github.io/ecma262/#sec-get-o-p">Get</a>(<var>prototype</var>, "attributeChangedCallback"). Rethrow any exceptions.</li>
+
+    <li>Let <var>definition</var> be a new <a>custom element definition</a> with <a href="#dfn-element-definition-type">custom element type</a> <var>type</var>, <a href="#dfn-element-definition-local-name">local name</a> <var>localName</var>, <a href="#dfn-element-definition-constructor">constructor</a> <var>constructor</var>, <a href="#dfn-element-definition-prototype">prototype</a> <var>prototype</var>, and <a href="#dfn-element-definition-lifecycle-callbacks">lifecycle callbacks</a> <var>attachedCallback</var>, <var>detachedCallback</var>, and <var>attributeChangedCallback</var>.</li>
+
+    <li>Add <var>definition</var> to <var>registry</var>.</li>
+
+    <li>Let <var>map</var> be <var>registry</var>'s <a>upgrade candidates map</a>.</li>
+
+    <li><a>Upgrade a newly-defined element</a> given <var>map</var> and <var>definition</var>.</li>
```
annevk, line 402: Shouldn't we also delete all the *Callback members at this point?

spec/custom/index.html
```diff
@@ -584,167 +459,229 @@
...
+<p>To allow creating both <a>custom tag</a> and <a>type extension</a>-style <a data-lt="custom element">custom elements</a>, the <a href="https://dom.spec.whatwg.org/#dom-document-createelement"><code>createElement</code></a> or <a href="https://dom.spec.whatwg.org/#dom-document-createelementns"><codE>createElementNS</code></a> methods gain optional <var>typeExtension</var> arguments. Their new definitions are:</p>
 
 <pre class='idl'>
 partial interface Document {
-    Element createElement(DOMString localName, DOMString typeExtension);
-    Element createElementNS(DOMString? namespace, DOMString qualifiedName, DOMString typeExtension);
+    Element createElement(DOMString localName, optional DOMString typeExtension);
+    Element createElementNS(DOMString? namespace, DOMString qualifiedName, optional DOMString typeExtension);
```
annevk, line 527: Should we go for a dictionary as final argument to allow future extensions without getting awkward four-argument methods? I think I'd prefer that.

spec/custom/index.html
```diff
@@ -584,167 +459,229 @@
...
-steps:</p>
 <ol>
-    <li>Let <var>TYPE</var> be <em>typeExtension</em>, or <em>localName</em> if <em>typeExtension</em> is not present</li>
-    <li>If an <a>element definition</a> with matching <em>localName</em>, <em>namespace</em>, and <var>TYPE</var> is <strong>not</strong> <a data-lt="element registration">registered</a> with <em>token</em>'s document, set <var>TYPE</var> to <em>localName</em></li>
-    <li>Let <em>interface</em> be the <a href="https://dom.spec.whatwg.org/#concept-element-interface">element interface</a> for <var>TYPE</var> as <a href="https://dom.spec.whatwg.org/#concept-element-local-name">local name</a> and <em>namespace</em> (<a href="http://www.w3.org/1999/xhtml">HTML Namespace</a> for <a href="https://dom.spec.whatwg.org/#dom-document-createelement"><code>createElement</code></a>)</li>
+    <li>If <var>localName</var> does not match the <a href="https://www.w3.org/TR/xml/#NT-Name"><code>Name</code></a> production, <a href="https://heycam.github.io/webidl/#dfn-throw">throw</a> an <code>InvalidCharacterError</code> exception.</li>
```
annevk: See #239.

spec/custom/index.html
domenic, line 402: No, that breaks calling `super.*Callback()` in subclasses.

spec/custom/index.html
domenic, line 527: Good idea, although I can't imagine Blink would be happy.

spec/custom/index.html
annevk, line 402: Oh right. It's still weird to me we store them (and call the stored versions) but also ask subclasses to call them directly without that kind of safety.

domenic:
OK, addressed everything except @annevk's idea of switching the typeExtension parameter to an options object. I will file a separate issue for that.

Maybe @hayatoito can give his signoff? Then I can squash into one commit and merge.

rniwa:
I think we should just merge what we have now and address additional editorial issues in separate PRs.

annevk:
Nice work @domenic!

WebReflection:
Apologies if this is not the best way to ask but I couldn't find the "end result" but a 404 here:
https://rawgit.com/w3c/webcomponents/construction/spec/custom/index.html

As maintainer of the only `document.registerElement` stand-alone polyfill, I'd like to know if the previous Custom Elements API has been already deprecated or if the new `defineElement` will simply coexist beside the good old API that both Polymer, X-Tags, and vanilla JS projects through my poly has been using for years now.

Thanks for any clarification and update.

annevk:
It's deprecated and non-standard. `defineElement()` is what all browsers will ship.

annevk:
You can find the specification here by the way: https://w3c.github.io/webcomponents/spec/custom/. It's been hosted there for a long time, not sure where you found the old URL.

Issue #431, opened 2016-03-11
Move defineElement elsewhere?


domenic:
Given the variety of proposals for new custom element related methods that are starting to fly around (e.g. #419 and #427 talk about upgrade-related helpers, plus there's always our plans for a registry API #154), we should start considering changes to the top-level API shape, and decide on something ASAP.

I can imagine three possibilities:
1. `document.customElements.define`. Everything goes on that object. `document.customElements` is null for documents without a browsing context.
2. `HTMLElement.define`. This makes it a bit clearer things are tied to the browsing context instead of the document. Other things can go there too. Right now the only things on there are a bunch of Node SHOUTY_CONSTANTS.
3. Punt on the problem: stick with `document.defineElement`, and just be sure to block future proposals on accepting one of the above two, at which point `document.defineElement` becomes an alias for one of those. This might not be feasible depending on how #419 goes.

@annevk @rniwa @travisleithead thoughts?

annevk:
`window.nodeRegistry.defineElement()`? I'm not sure why we'd hang it off `document`. I don't think putting it on `HTMLElement` is future-proof. And we might introduce other custom nodes in the future, maybe?

domenic:
I don't think we're going to generalize this to arbitrary node types :). But yeah, we might allow custom SVG or MathML elements, and I suppose making people decide between `SVGElement.define` and `HTMLElement.define` would be bad?

`window.customElements` seems OK I guess.

rniwa:
Or `window.customElementDefinitions` since `window.customElements` sounds like a list that contains instances of custom elements.

domenic:
I think in context it's fine. `window.customElements.define('...')` etc.

JanMiksovsky:
FWIW, I've occasionally wished for a way to know what custom elements have been registered.
- When debugging a complex component-based app with multiple levels of module dependencies and other complications, it can be helpful to just get confirmation that a particular custom element has, in fact, being successfully registered.
- I've already hit cases where I wanted to provide a definition for a component if a definition hadn't already been registered. It's possible to just call createElement('foo-bar') and sniff the result, but it'd be nice to have a more deterministic way.
- Someday I'd like the debugger to be able list the currently-registered elements, let me jump to their declarations, etc.

While @rniwa didn't intend to suggest a feature, I actually think having window.customElements provide a list of registered elements would be pretty interesting. It could be a dictionary or Map of element names to classes, for example.

```
class FooBar extends HTMLElement {...}
[...]

PR #446, opened 2016-03-17
Fix typo in '2.1.1 Creating a custom tag'

rianby64:
According to #431 the example given in the documentation should use _document.registerElement_ instead of _document.defineElement_

### DIFF
--- a/spec/custom/index.html
+++ b/spec/custom/index.html
@@ -176,7 +176,7 @@ class FlagIcon extends HTMLElement {
 <p>We then need to use this class to define the element:</p>
 
 <pre class="highlight">
-document.defineElement("flag-icon", FlagIcon);</pre>
+document.registerElement("flag-icon", FlagIcon);</pre>
 
 <p>At this point, our above code will work! The parser, whenever it sees the <code>flag-icon</code> tag, will construct a new instance of our <code>FlagIcon</code> class, and tell our code about its new <code>country</code> attribute, which we then use to set the element's internal state and update its rendering (when appropriate).</p>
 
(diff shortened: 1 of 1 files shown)


spec/custom/index.html
```diff
@@ -176,7 +176,7 @@ <h4 id="custom-elements-custom-tag-example">Creating a custom tag</h4>
 <p>We then need to use this class to define the element:</p>
 
 <pre class="highlight">
-document.defineElement("flag-icon", FlagIcon);</pre>
+document.registerElement("flag-icon", FlagIcon);</pre>
```
rniwa, line 179: This isn't right.  We should be using `window.customElements.define` instead.

rianby64, line 179: Agree. But if you check window.customElements in browser then it isn't define yet. But, document.registerElement at least is present in Firefox and Chrome. For the shake of the example let's consider use registerElement. After all by using registerElement the example will work in some way.

andyearnshaw, line 179: The example should reflect the specification, not the current implementations in browsers.

rianby64, line 179: Yes, you're right! [This interface 2.4 The CustomElementsRegistry interface](http://w3c.github.io/webcomponents/spec/custom/#custom-elements-api) should be used.

Anyway. Will be great if other developers will read an example that works. The introduction _section is non-normative_ after all. So, today we can read an example that works and allows us to understand the meaning of custom components.

domenic:
Thanks @rianby64! Updated in d0f7fec93f9705968d5ba8c2197bf79f3f1db3b2 to use the correct `customElements.define` API.

PR #447, opened 2016-03-18
Proposed change for Focus Navigation

TakayoshiKochi:
This is updated patch for #381, to address issue #375.

The previous branch went somewhat weird state and recreated a branch.


TakayoshiKochi:
@hayatoito PTAL
This is a continuation from #381

hayatoito:
At the first glance, it looks good to me. I'll take another deeper look later.

Given that this is the formal definition, I am afraid that it is not easy to understand how this works actually.
It would be great that we have a simple example.

spec/shadow/index.html
```diff
@@ -1149,21 +1149,150 @@
       </section>
 
       <section>
-        <h3>Focus Navigation</h3>
+        <h3>Sequential Focus Navigation</h3>
 
-        <p>If a <a>node</a> doesn’t <a data-lt="participates">participate</a> in the <a>document flat tree</a>, the <a>node</a> <strong>must</strong> be skipped from the <a>sequential focus navigation</a>.</p>
+        <p><dfn>focus navigation scope</dfn> is a set of elements <a>in a composed document</a>. An element belongs to at most one <a>focus navigation scope</a>.</p>
```
hayatoito: Could you avoid to use _composed document_ in normative section?

spec/shadow/index.html
```diff
@@ -1149,21 +1149,150 @@
...
+        <div class="algorithm">
+          <dl>
+            <dt>INPUT</dt>
+            <dd><var>DOCUMENT</var>, a composed document</dd>
+            <dt>OUTPUT</dt>
+            <dd><var>OWNERS</var>, a set of <a>focus navigation scope owner</a> nodes</dd>
+          </dl>
+          <ol>
+            <li>Let <var>OWNERS</var> be an empty set.</li>
+            <li>For each node tree <var>TREE</var> in <var>DOCUMENT</var>:
```
hayatoito, line 1140: Now you can use https://dom.spec.whatwg.org/#concept-shadow-including-tree-order.
Or, given that the order is not important, you can simply say:
- Let OWNERS be focus owner nodes whose shadow-including root is _document_.

TakayoshiKochi:
@hayatoito PTAL

hayatoito:
I think we can merge this now.
I think it's time to edit this on HEAD. We can improve this anytime.

PR #448, opened 2016-03-18
Fix typo in "2.1.1" for attributeChangedCallback

rianby64:
The example given at "2.1.1 Creating a custom tag" shows how to use

attributeChangedCallback(name, newValue)

but "4.2 DOM+: Mutation algorithms" last paragraph states that
this function has 3 arguments (name, oldAttrValue, newAttrValue)
so, newValue given in the example doesn't reflect the meaning of
oldAttrValue


domenic:
Thank you!

PR #456, opened 2016-03-23
Fix various typos

pimterry:
Some quick typo fixes:
- Custom **c**lements -> Custom elements
- CustomElementRegistry -> CustomElement**s**Registry (both are used interchangably, but the latter much more commonly, and it matches its `window.customElements` property name.


domenic:
Thanks! Merged as ed336665e13c096eb0fd28c14582599eedfa3623.

PR #498, opened 2016-05-10
add a script for auto-publication 

siusin:
Fyi, this script will help get the WebComponents specs under /TR to sync with the latest commit in GitHub.

(Many thanks to @deniak @plehegar , please see "[Using Echidna](https://github.com/w3c/echidna/wiki/How-to-use-Echidna)" for more information about setting up the auto publication system.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: node_js
+sudo: false
+branches:
+  only:
+  - gh-pages
+env:
+  global:
+  - URL=""
+  - DECISION="https://lists.w3.org/Archives/Public/public-webapps/2014JulSep/0627.html"
+  - secure: "IOj5spID70zoohkd8SpllAaZUtquHV2gnlhSxpBJWfKhuB17KaUAMB4gD7RBn/jn4jv1iHUg6h27gVn5YSOTzaLcFDZpRcbrdj5Jeq9syBYZzGsyno3vkjxB8wsbFuKOpJjRBRyANlqeyONxpy4q7/PGQzm3Xb9WARr/Tp8BU4c="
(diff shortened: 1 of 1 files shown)


rniwa:
@hayatoito @domenic

domenic:
I don't really know how to review this, so feel free to merge if you think it will work. Auto-publishing sounds like a good idea.

plehegar:
It's missing "$TRAVIS_PULL_REQUEST = false &&" on each test line, otherwise you'll trigger the publication on every pull request instead of just commits into the main branch (ie you'll generate TR updates that needed).

Otherwise it looks reasonable.

domenic:
I think the `branches: only: gh-pages` is meant to take care of that...

plehegar:
that wasn't enough unfortunately. we used to generate a lot more TR drafts for webperf until I added this.
btw, I just tried to manually trigger the publication [1] but /assets/styles/respec-complement.css is getting in the way. index.html expects that css to be in "../../assets/styles/respec-complement.css" but that won't match the location at the end.
Assuming the file is still relevant, we could:
1. move a copy of respec-complement.css into spec/custom/ but we'll have to do the equivalent for shadow dom
2. auto-include the css if specStatus != WD using a respec pre or post script

any preference or got a different idea?

[1] https://lists.w3.org/Archives/Public/public-tr-notifications/2016May/0031.html

domenic:
We could maintain separate copies for each spec. Everything above line 53 is not needed by custom elements, and lines 80-96 are not needed by specs that are not custom elements.

I can work on this but would need a way to locally test whether publication is going to work. Happy to have you help if you think that would be easier.

plehegar:
working on it.

siusin:
Script updated - putting "$TRAVIS_PULL_REQUEST = false &&" into the test lines.

deniak:
@siusin it should be
`- test $TRAVIS_PULL_REQUEST = false && test "${SPECS#*custom}" != "$SPECS"`
instead of 
`- test $TRAVIS_PULL_REQUEST = false && "${SPECS#*custom}" != "$SPECS"`

siusin:
@deniak Thank you! - updated now.

Issue #538, opened 2016-08-09
Broken wikipedia link for "Green Eye"


jfw7:
https://w3c.github.io/webcomponents/spec/shadow/#shadow-dom-example

"After getting another cup of **[Green Eye](https://en.wikipedia.org/wiki/List_of_coffee_drinks)**, he quickly mocks up the following shadow tree, to be hosted by the ul element"

"Green Eye" appears to have been removed from "List of coffee drinks" in January 2015.

hayatoito:
WONTFIX as of now.

PR #539, opened 2016-08-09
Change Bob's coffee drink from "green eye" to "red eye"

jfw7:
Potential fix for #538.  A "green eye" is a Starbucks beverage with 3 shots of espresso added to coffee, so a "red eye" or another drink with [coffee and espresso](https://en.wikipedia.org/wiki/List_of_coffee_drinks#Coffee_with_espresso) would be the closest replacement.


hayatoito:
This is not a correct fix because Bob loves Green Eye. We should update the Wikipedia for Bob, then update the link. Could you rescue Bob?

PR #557, opened 2016-08-29
Fix #556, make focus navigation work when shadow host's tabindex is -1

### LINKED ISSUE #556: Shadow host with its .tabindex property -1 would be skipped for focus navigation
Context:
https://github.com/w3c/webcomponents/issues/554#issuecomment-242636505

In the Shadow DOM spec 6.3
http://w3c.github.io/webcomponents/spec/shadow/#sequential-focus-navigation
specifically in Step 3.1.3, all elements with negative tabindex value will be skipped in the focus navigation, but shadow hosts with `delegatesFocus=faluse` should be excluded otherwise all focusable elements in its shadow root will never visited.

TakayoshiKochi:
The Shadow DOM spec section 6.3 (sequential focus navigation) unintentionally excluded all shadow hosts with nagative tabindex in Step 3.1.3. This was unexpected for shadow hosts with `delagatesFocus=false` shadow root.

Before:
- Exclude all shadow hosts with tabindex < 0

After:
- Allow shadow hosts with tabindex < 0, but only when `delegatesFocus=false`
- In sorting by tabindex, treat such shadow hosts as `tabindex=0`
- Exclude those shadow hosts in Step 4.

### DIFF
--- a/spec/shadow/index.html
+++ b/spec/shadow/index.html
@@ -1153,7 +1153,7 @@
             <li>When <var>HOST</var> is focused by <code>focus()</code> method or <code>autofocus</code> attribute: The first <a>focusable area</a> in focus navigation order of <var>HOST</var>'s <a>shadow root</a>'s <a>focus navigation scope</a> gets focus. See the next section for the formal definition of the ordering.</li>
             <li>When mouse is clicked on a node in <var>HOST</var>'s <a>shadow tree</a> and the node is not a <a>focusable area</a>: The first <a>focusable area</a> in focus navigation order of <var>HOST</var>'s <a>shadow root</a>'s <a>focus navigation scope</a> gets focus. See the next section for the formal definition of the ordering.</li>
             <li>When any element in <var>HOST</var>'s <a>shadow tree</a> has focus, <code>:focus</code> pseudo-class applies to <var>HOST</var> in addition to the focused element itself.</li>
-	    <li>If <code>:focus</code> pseudo-class applies to <var>HOST</var>, and <var>HOST</var> is in a <a>shadow root</a> of another <a>shadow host</a> <var>HOST2</var> which also <a>delegates focus</a>, <code>:focus</code> pseudo-class applies to <var>HOST2</var> as well.</li>
+            <li>If <code>:focus</code> pseudo-class applies to <var>HOST</var>, and <var>HOST</var> is in a <a>shadow root</a> of another <a>shadow host</a> <var>HOST2</var> which also <a>delegates focus</a>, <code>:focus</code> pseudo-class applies to <var>HOST2</var> as well.</li>
           </ol>
         </p>
       </section>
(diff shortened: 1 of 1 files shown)


spec/shadow/index.html
```diff
@@ -1260,9 +1260,15 @@
...
                 <li>For each element <var>ELEMENT</var> in a <a>focus navigation scope</a> <var>SCOPE</var>,
                   <ol>
-                    <li>if <var>ELEMENT</var> is focusable, a <a>shadow host</a>, or a <a>slot element</a>, append <var>ELEMENT</var> to <var>NAVIGATION-ORDER</var>.</li>
+                    <li>If <var>ELEMENT</var> is a <a>shadow host</a>:
+                      <ol>
+                        <li>If its <a>tabindex</a> value is not negative and its <a>shadow root</a>'s delegatesFocus flag is set, append <var>ELEMENT</var> to <var>NAVIGATION-ORDER</var>.</li>
+                        <li>Otherwise, append <var>ELEMENT</var> to <var>NAVIGATION-ORDER</var>.</li>
+                      </ol>
+                    </li>
+                    <li>Otherwise if <var>ELEMENT</var> is focusable, or a <a>slot element</a>, and its <a>tabindex</a> value is not negative, append <var>ELEMENT</var> to <var>NAVIGATION-ORDER</var>.</li>
```
hayatoito, line 1269: The "and", "or" relationship are confusing. Could you make this explicit?

spec/shadow/index.html
```diff
@@ -1260,9 +1260,15 @@
                 <li>Let <var>NAVIGATION-ORDER</var> be an empty list.</li>
                 <li>For each element <var>ELEMENT</var> in a <a>focus navigation scope</a> <var>SCOPE</var>,
                   <ol>
-                    <li>if <var>ELEMENT</var> is focusable, a <a>shadow host</a>, or a <a>slot element</a>, append <var>ELEMENT</var> to <var>NAVIGATION-ORDER</var>.</li>
+                    <li>If <var>ELEMENT</var> is a <a>shadow host</a>:
+                      <ol>
+                        <li>If its <a>tabindex</a> value is not negative and its <a>shadow root</a>'s delegatesFocus flag is set, append <var>ELEMENT</var> to <var>NAVIGATION-ORDER</var>.</li>
+                        <li>Otherwise, append <var>ELEMENT</var> to <var>NAVIGATION-ORDER</var>.</li>
```
hayatoito, line 1266: It looks a shadow host is always appended to NAVIGATION-ORDER. :(

spec/shadow/index.html
```diff
@@ -1260,9 +1260,15 @@
...
+                      </ol>
+                    </li>
+                    <li>Otherwise if <var>ELEMENT</var> is focusable, or a <a>slot element</a>, and its <a>tabindex</a> value is not negative, append <var>ELEMENT</var> to <var>NAVIGATION-ORDER</var>.</li>
                   </ol>
-                <li>Reorder <var>NAVIGATION-ORDER</var> according to the <a>tabindex</a> value attached to each node. In this step, an element whose <a>tabindex</a> value is negative <strong>must</strong> not be appended to <var>NAVIGATION-ORDER</var>.</li>
+                <li>Reorder <var>NAVIGATION-ORDER</var> according to the <a>tabindex</a> value attached to each node. In this step, a shadow host with negative <a>tabindex</a> value must be treated as if its <a>tabindex</a> value were 0.</li>
```
hayatoito, line 1271: Could you have a link which explains how "Reorder" sort elements?

In the current definition, it looks that elements would be sorted by only "tabindex". No tie-breaker.
We should have a link to explain how these elements are sorted by tabindex and tree-order.

spec/shadow/index.html
```diff
@@ -1284,7 +1290,7 @@
               <ol>
                 <li>If <var>ELEMENT</var> is a shadow host:
                   <ol>
-                    <li>Unless its shadow root’s delegatesFocus flag is set, append<var> ELEMENT</var> to <var>MERGED-NAVIGATION-ORDER</var>.</li>
+                    <li>If it is focusable, its <a>tabindex</a> value is not negative, and its <a>shadow root</a>’s delegatesFocus flag is not set, append <var>ELEMENT</var> to <var>MERGED-NAVIGATION-ORDER</var>.</li>
```
hayatoito, line 1293: This condition in the sentence might be confusing. Could you say ELEMENT has to satisfy every conditions explicitly?

spec/shadow/index.html
TakayoshiKochi: This is intentional.  Unless `delegatesFocus` flag is set, put the shadow host in the NAVIGATION-ORDER regardless of the tabindex value, to mark the point into which its shadow scope will be merged at the next step.

hayatoito: No. My point is:

```
If (ELEMENT is a shadow host) {
  If (... some condition...) {
    Add appendELEMENT to NAVIGATION-ORDER.
  } else {
    Add appendELEMENT to NAVIGATION-ORDER.
  }
}    
```

This is equivalent to:

```
If (ELEMENT is a shadow host) {
  Add appendELEMENT toNAVIGATION-ORDER.
}    
```

Please re-check the condition and wording.

Donsdon (approved):

PR #580, opened 2016-10-02
[minor] typo in interface name

rictic:
The interface is defined as CustomElementRegistry not CustomElementsRegistry; looks like a typo.


siusin:

PR #586, opened 2016-10-12
Use NodeList.prototype.forEach

shvaikalesh:


siusin:

PR #632, opened 2017-03-24
fix example so that attachShadow does not throw an exception

dylanb:
An unordered list `<ul>` is one of the elements that does not allow `attachShadow()` to be called. The example tries to call `attachShadow()` on the `<ul>`. This commit fixes this problem.


hayatoito:
Thanks. However, it looks that this example needs more fix because the shadow host's children are `<li>`. I am afraid that it does not make sense to use `<li>` here if we do not use `<ul>` as a shadow host..

And the HTML Standard does not guarantee that `<ul><slot></slot></ul>` work at all. Even if `<li>` is assigned to a slot, we can not say that `<ul>` and `<li>` are in *parent-child* relationship here. This issue is still in the air.

Westbrook:
Out of date.

PR #633, opened 2017-03-28
Fix typo: change naive to native

comfroels:


TakayoshiKochi:
I don't think this is a typo, it is intentionally "naive". (or "naïve", maybe?).

PR #687, opened 2017-11-02
Copy edits for all remaining parts

sideshowbarker:


proposals/Template-Instantiation.md
```diff
@@ -298,9 +298,9 @@ document.defineTemplateType('with-for-each', { processCallback: (instance, parts
 list.createInstance({items: [{class: 'baz', value: 'baz', label: 'hello world'}]});
 ```
 
-In fact, when `update` method of `TemplateInstance` is subsequently called, we might need to create more template parts and remove some since the number of elements in the array for looping (“items” in this case) can change from one update to another.
+In fact, when the `update` method of `TemplateInstance` is subsequently called, we might need to create more template parts and remove some, since the number of elements in the array for looping (“items” in this case) can change from one update to another.
 
-As suggested by others in the past discussions, we propose to use nested templates for these cases:
+As suggested by others in past discussions,fo these cases  we propose to use nested templates:
```
rniwa, line 303: What happened here?

proposals/Template-Instantiation.md
```diff
@@ -338,7 +338,7 @@ To make this process more streamlined, we propose treating inner template elemen
 };
 ```
 
-With this interface, the template process callback that implements foreach and other kinds of looping constructs could simply call `replace` with newly constructed elements as follows:
+With this interface, the template process callback that implements a *foreach* and other kinds of looping constructs could simply be called `replace` with newly constructed elements, as follows:
```
rniwa, line 341: The original sentence was correct. "The template process callback ~ could simply call `replace` with newly constructed elements,".

proposals/Template-Instantiation.md
```diff
@@ -356,36 +356,33 @@ document.defineTemplateType("with-for-each", {
...
 
-There is an alternative approach for (9), which is to make TemplatePart constructible from template process callback themselves. However, this approach involves the template process callback cloning template parts along with other nodes, or let author scripts manually specify to which element each template part belongs. This quickly becomes an entangled mess because now we could have multiple template parts that refer to a single DOM location or an attribute, and we have to start dealing with multiple template parts trying to override one another even though there is no good use case for such a behavior.
-
-We like the idea of supporting very basic control flow such as `if` and `foreach` in the default template process callback but we don't think it's a show stopper if the default template process callback didn't support them in the initial cut.
-
+We like the idea of supporting very a basic control flow such as `if` and `foreach` in the default template process callback, but we don't think it'd be a show stopper if the default template process callback didn't support them in the initial cut.
```
rniwa, line 363: This should have said "supporting very basic control flow statements".

proposals/Template-Instantiation.md
```diff
@@ -395,45 +392,45 @@ The  `createInstance(optional any state)` method on `HTMLTemplateElement`, when
...
-        4. Append *innerPart* to the end of *parts.*
-        5. [Remove](https://dom.spec.whatwg.org/#concept-node-remove) *currentNode* from the *currentNode*'s [parent](https://dom.spec.whatwg.org/#concept-tree-parent).
-        6. Run the concept to _apply node template part list_ with *nodeValueSetter*.
+5. For every [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) node *currentNode* of *instance* in [tree order](https://dom.spec.whatwg.org/#concept-tree-order), run these steps:
+    1. If *currentNode* is a [`template` element](https://html.spec.whatwg.org/multipage/scripting.html#the-template-element):
+      1. Run the concept to _adjust single node case_ with *currentNode*.
+      2. Let *nodeValueSetter* be a new instance of the _node value setter_ with *currentNode*, the [previous sibling](https://dom.spec.whatwg.org/#concept-tree-next-sibling) of *currentNode*, the [next sibling](https://dom.spec.whatwg.org/#concept-tree-next-sibling) of *currentNode*, an empty _previous replacement nodes_, _fully templatized_ flag set to the result of running the concept to _determine full templatizability_ with *currentNode*, and an empty _node template part list_.
```
rniwa, line 398: I'm using _ to only refer to variable names. So please don't italicize concepts defined elsewhere.

proposals/Template-Instantiation.md
```diff
@@ -395,45 +392,45 @@ The  `createInstance(optional any state)` method on `HTMLTemplateElement`, when
...
-        3. Let *innerPart* be a new instance of `InnerTemplatePart` associated with *currentNode*, an empty _replacement node list_, and *nodeValueSetter*.
-        4. Append *innerPart* to the end of *parts.*
-        5. [Remove](https://dom.spec.whatwg.org/#concept-node-remove) *currentNode* from the *currentNode*'s [parent](https://dom.spec.whatwg.org/#concept-tree-parent).
-        6. Run the concept to _apply node template part list_ with *nodeValueSetter*.
+5. For every [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) node *currentNode* of *instance* in [tree order](https://dom.spec.whatwg.org/#concept-tree-order), run these steps:
+    1. If *currentNode* is a [`template` element](https://html.spec.whatwg.org/multipage/scripting.html#the-template-element):
```
rniwa, line 396: I'd like to keep using 4-space indentations. Please revert.

PR #691, opened 2017-11-02
Fix Markdown styles of Template-Instantiation

koba04:


koba04:

PR #696, opened 2017-11-04
Replace Em dash with hyphen minus

koba04:
This can be hard to find the difference on markdown...

screen shot 2017-11-05 at 0 56 08


koba04:

koba04:
Thank you!

PR #698, opened 2017-11-06
Add a missing parenthesis

koba04:


koba04:

PR #706, opened 2017-11-13
Fix an arguments order of replaceChild

koba04:
> replacedNode = parentNode.replaceChild(newChild, oldChild);

* https://developer.mozilla.org/en-US/docs/Web/API/Node/replaceChild


koba04:

Issue #783, opened 2019-01-17
HTML modules spec draft editorial feedback


domenic:
Hey @dandclark, @bocupp, @samsebree, @travisleithead. Excited to see the blink-dev intent to implement! I wanted to provide some spec feedback on https://github.com/w3c/webcomponents/blob/gh-pages/proposals/html-module-spec-changes.md in the hopes of making the future integration into HTML smoother. It's all generally editorial, but I thought getting it out there sooner would be a good idea.

- We should define the new module record type in the HTML spec, not the ES spec. ES provides a base class, which gets specialized by different host environments which have their own needs---e.g. the WebAssembly spec will have its own type of module record, and HTML modules should have theirs. We shouldn't bake HTML-specific semantics into the core language.
- We should rebase the spec on top of https://github.com/tc39/ecma262/pull/1311. I am not sure why that hasn't landed yet, but I've [asked](https://github.com/tc39/ecma262/pull/1311#issuecomment-455039992). That should make things easier in various ways.
- Since HTML modules will continue to be included through `<script type="module">`, we should continue using the script infrastructure. As such, just as a naming thing, the spec-level struct should be "HTML module script", which is a third type of [script](https://html.spec.whatwg.org/multipage/webappapis.html#concept-script). We should probably then rename the existing "module script" type to "JS module script"; that will be needed anyway when we introduce WebAssembly module scripts.
- [[ScriptName]] is a confusing slot name; I'd suggest [[ExternalScriptURL]]. Then maybe [[ModuleRecord]] could be named [[InlineModuleRecord]]. You could also probably get rid of [[IsInline]], just keying off the null-ness or not of those other two slots.

Overall, very exciting stuff!

dandclark:
This is all great feedback, thanks!  I'll work on getting these changes incorporated into the docs.

annevk:
Off-topic: @dandclark there's also other issues labeled "modules" with feedback that doesn't appear addressed (e.g., the suggested MIME type is `text/html` iiuc).

dandclark:
Thanks @annevk -- I'll make sure we get to these too.

dandclark:
@domenic,

For your third bullet point, what do you think about HTML Module Script/JS Module Script (and eventually JSON Module Script etc) types sharing a common abstract Module Script base type?
That is, do we want type hierarchy #&#x2060;1:
Classic Script is a Script
Module Script is a Script
JS Module Script is a Module Script
HTML Module Script is a Module Script

[...]

Issue #742, opened 2018-03-06
HTML module should require a new MIME type


annevk:
It seems navigating directly to an HTML module should not result in rendering and script execution. A download seems safer. (Need to make sure it doesn't get sniffed, either.)

rniwa:
`text/html-module` or `application/html-module` perhaps? I guess we'd need to make a recommendation for a new file extension as well. e.g. `mymodule.htmlmodule`. If regular HTML files and HTML modules both end with  `.html` or `.htm`, it's going to be really annoying to give a different MIME type in various server environment as well as local environment. e.g. `file` on macOS and Windows probably won't be able to get the right MIME type.

tomalec:
I'd like to give another use case for using specific mime types.
Currently, I'm using HTML Imports to build client-side includes driving SPA. I'd like to use HTML Modules for this as well.

I would like to be able to request the same URL to open a new complete page (`text/html`) as one to request a module with the partial view and dependencies (`text/html+module`) to morph existing page into.

ExE-Boss:
If anything, it should probably be `application/html-module+html` or `application/xhtml-module+xml`.

(The content after the `+` signifies encoding, so `text/html+module` would violate the MIME type specification, but `text/module+html` or `application/module+html` wouldn’t)

---

Another issue, at least when using HTML modules inside a XHTML document, is that XML determines the element type based on the namespace, which is what #752 intends to solve.

annevk:
I don't think we should use `+` here since we don't want to parse everything as `+html` as HTML (as we do with `+xml`) and we don't want to allow navigating to these resources either. Navigating to them should download them or result in some special view.

`text/html-module` and `.htmlm` (after `.jsm`) seem most reasonable.

dandclark:
I still want to question whether this is a necessary change to make.

@annevk said: 
> "It seems navigating directly to an HTML module should not result in rendering and script execution. A download seems safer."

Is there a security issue?  What is the threat?

If you want the module to be downloaded when navigated to directly, it could be served with a content-disposition header, which will be ignored when loaded as a module.

If you want to differentiate an HTML page's behavior based on whether it's loaded as an HTML module or stand-alone page, you can feature detect whether you are executing a script module as part of an HTML module using `import.meta.document`.  Maybe there are use cases like displaying a module's documentation if it's requested as a top-level page.

In terms of prior art… fetching module scripts today requires that the mime-type match the existing JS mime-types.  Why does an HTML module need a new one?
[...]

PR #793, opened 2019-02-13
Update HTML Modules Proposed Spec changes and initial proposal document

dandclark:
I've made significant updates to the HTML Module proposed spec changes doc based on feedback [here](https://github.com/w3c/webcomponents/issues/783) and [here](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/ewfRSdqcOd8).

Highlights include:

•	Rebased the ES changes on https://github.com/tc39/ecma262/pull/1311.  The most obvious change here is that HTML Module Record is now a subclass of Cyclic Module Record.
•	HTML MR now inherits the concrete implementations of Instantiate() and Evaluate() from Cyclic MR (HTML MR still defines its own InnerModuleInstantiation /InnerModuleEvaluation.
•	Updated ResolveExport to return a ResolvedBinding rather than the document itself.  HTML MR implementations of InitializeEnvironment() and ExecuteModule() were added to create a binding named “*default*” and set it to the document.
•	Corrected GetExportedNames to refer to HTML MR instead of Source Text MR.
•	HTML Module on the HTML5 side is now a type of Script called HTML Module Script, and the existing Module Script is renamed to JavaScript Module Script.
•	ScriptEntry’s [[ScriptName]] field is renamed to [[ExternalScriptURL]] and [[ModuleRecord]] is renamed to [[InlineModuleRecord]].  The [[IsInline]] field was removed.
•	Fixed ResolveExport so that it the HTML Module’s default export doesn’t override a default export specified by an inline script.
•	Updated and refactored “create a module script” and ParseHTMLModule to ensure that all fields are initialized for HTML Module Script and HTML MR creation.
•	Edited HTML Module proposal doc to state that non-module scripts in an HTML Module will cause a parse error and make the module to fail to create, rather than being silently coerced to `type="module`.

There are still some questions around whether the bulk of HTML Module Record and its algorithms should actually be defined in the HTML5 spec.  I expect this document to continue to evolve.

### DIFF
--- a/proposals/html-module-spec-changes.md
+++ b/proposals/html-module-spec-changes.md
@@ -1,4 +1,4 @@
-This is a list of spec areas that will need to be changed to implement our [HTML Modules proposal](https://github.com/w3c/webcomponents/blob/gh-pages/proposals/html-modules-proposal.md).  Some of the proposed changes are very specific and some are more handwavy depending on how deep I've investigated into that particular area.
+This is a list of spec areas that will need to be changed to implement our [HTML Modules proposal](https://github.com/w3c/webcomponents/blob/gh-pages/proposals/html-modules-proposal.md)
 Questions/corrections/feedback are welcome!  I've left TODOs in several places where we still have open questions; any input regarding these is especially appreciated.
 
 -- [@dandclark](https://github.com/dandclark), with:\
@@ -15,8 +15,8 @@ These spec changes are built on top of the proposed refactoring here: https://gi
     
     | Field Name | Value Type | Meaning |
     | --- | --- | --- |
-    | [[InlineModuleRecord]] | Module Record \| null | The Module Record for the module request.  May be set when the module is created, or initially Null and set later during module Instantiation.  |
-    | [[ExternalScriptURL]] | String \| null | The URL for the module request if the Module Record was not available at module creation time.  Null otherwise. |
+    | [[InlineModuleRecord]] | Module Record \| null | The Module Record for the module request if available at ScriptEntry creation time.  Null otherwise.  |
+    | [[ExternalScriptURL]] | String \| null | The URL for the module request if the Module Record was not available at ScriptEntry creation time.  Null otherwise. |
     2. The [[HostDefined]] field in Abstract Module Record will be set to the HTML Module Script (see HTML5 Spec changes).  This is analogous to script modules where this field holds the JavaScript [module script](https://html.spec.whatwg.org/multipage/webappapis.html#module-script) as defined in the HTML5 spec.
 - HTML Module Record inherits the concrete Instantiate() method from Cyclic Module Record.
 - HTML Module Record defines its own version of [InnerModuleInstantiation](https://tc39.github.io/ecma262/#sec-innermoduleinstantiation).  Cyclic Module Record's definition recursively calls InnerModuleInstantiation on each child module (calling [HostResolveImportedModule](https://tc39.github.io/ecma262/#sec-hostresolveimportedmodule) to resolve module names to Module Records), then calls [ModuleDeclarationEnvironmentSetup](https://tc39.github.io/ecma262/#sec-moduledeclarationenvironmentsetup) to set up the lexical environment and resolve imports/exports for the current module.  HTML Module Record's version will be similar, but will change the definition of step 9 (“For each string required that is an element of module.[[RequestedModules]]...) in order to follow the structure of the newly defined ScriptEntry record as defined above.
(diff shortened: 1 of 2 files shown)


proposals/html-module-spec-changes.md
```diff
@@ -52,65 +71,89 @@ Questions/corrections/feedback are welcome!  I've left TODOs in several places w
...
-  - 1\. Let *record* be a new HTML Module record that this algorithm will subsequently initialize. 
-  - 2\.	Run the HTML5 parser on source to obtain the result *document*.
-  - 3\.	Set *record*.[[HostDefined]] = *htmlModule*.
-  - 4\.	For each HTMLScriptElement *script* in *document*:
-    - a\.	Let *se* be a new ScriptEntry record (see definition in ES6 changes above).
+      - ii\.	Set *htmlModuleScript*'s *parse error* to *error*.
+      - iii\.	Return *htmlModuleScript*.  Note: This step is essentially validating all of the requested module specifiers. We treat a module with unresolvable module specifiers the same as one that cannot be parsed; in both cases, a syntactic issue makes it impossible to ever contemplate instantiating the module later.
+  - 8\. Set *htmlModuleScript*'s *record* to *result*.
+  - 9\. Return *htmlModuleScript*.
+- Introduce a new algorithm ParseHTMLModule(*source*, *realm*, *htmlModuleScript*) as the following.
```
annevk, line 122: I think this needs to be more detailed. In particular what I think we want to do is to run the parser in a mode similar to XMLHttpRequest, whereby nothing is executed or fetched. And then once done we walk the resulting tree and execute scripts provided they meet the requirements.

It also seems reasonable to me to not require the `type` attribute as this is a different processing model so we can also offer developer conveniences.

proposals/html-module-spec-changes.md
```diff
@@ -52,65 +71,89 @@ Questions/corrections/feedback are welcome!  I've left TODOs in several places w
...
+      - ii\. Set *se*[[ExternalScriptURL]] = *script’s* src URL
+    - d\.	Append *se* to *scriptEntries*.
+  - 4\. Return a new HTML Module Record { [[Realm]]: *realm*, [[Environment]]: *undefined*, [[Namespace]]: *undefined*, [[Status]]: `"uninstantiated"`, [[EvaluationError]]: *undefined*, [[HostDefined]]: *htmlModuleScript*, [[RequestedModules]]: *scriptEntries*, [[DFSIndex]]: *undefined*, [[DFSAncestorIndex]]: *undefined* }.
+- Provide an implementation of the abstract operation HostGetDefaultExport(*module*) as the following:
+  1. Assert: *module* is an HTML Module.  Note: It is anticipated that in the future this may be expanded to support other module types (JSON etc).
+  1. Let *htmlModuleScript* be *module*.[[HostDefined]].
+  1. Let *document* be *htmlModuleScript*'s *document*.
+  1. Return *document*.
+- Change [fetch a single module script](https://html.spec.whatwg.org/#fetch-a-single-module-script) as follows:
+  - In step 9, allow `text/html` type through in addition to JavaScript.
```
annevk: As per #742 I object to this.

dandclark: Yeah, that's fair.  I changed the language to reflect that the MIME type discussion is still ongoing, and linked [#742](https://github.com/w3c/webcomponents/issues/742).

proposals/html-module-spec-changes.md
dandclark: Agreed that this needs to be fleshed out more.  Script execution should already be [disabled](https://html.spec.whatwg.org/#concept-n-noscript) because the module doc doesn't have a browsing context, but the case for fetches isn't so clear.  Most things (``, `<iframe>`, etc) won't kick off fetches until moved to the main document, but it's possible something I'm not thinking of could slip through and perhaps we should head it off by enforcing a special parser mode.  I'd prefer not to block all the other changes on this though so I added a TODO to investigate/deliberate this further.

The "walking through the resulting tree and executing scripts" process is already handled; see the "For each HTMLScriptElement *script* in *document*" part where we feed the scripts into the ES Modules infra (as the ScriptEntries of the HTML Module Record), so they'll be executed as part of the module graph's Evaluation phase.

Our original proposal had it such that we'd coerce non-module type scripts to module-type, but that got pushback from a few differnent folks that it would be confusing to devs and we should just make non-module scripts an error.  I'm not totally sold on either direction; do you think it's worth opening a new issue thread to discuss this specifically?

annevk: Sounds good to me. I don't think it'll be confusing long term as HTML modules are quite distinct from HTML documents in many respects.

annevk: (And thanks for explaining the way script execution happens. That makes sense I think.)

travisleithead:
Thanks Dan! (We just got the IPR block sorted...)

proposals/html-module-spec-changes.md
dandclark: I belatedly opened an issue [here](https://github.com/WICG/html-modules/issues/2) for the question of whether to convert classic scripts to module scripts or to throw an error.

It's in the new WICG/html-modules repo for now but I'll move it if we end up not using that.

annevk: I think we'd make faster progress here. There's 290 people watching this repository. There's 6 people watching that.

dandclark: Yup, we moved it to https://github.com/w3c/webcomponents/issues/798.

Edit: I guess you saw it already. :)

PR #844, opened 2019-09-30
Custom state: Preparation for TAG review

tkent-google:
Add "Security and Privacy Considerations", "Major open issues", "Discussions", and "Self-Review Questionnare on Security and Privacy" sections

### DIFF
--- a/proposals/custom-states-and-state-pseudo-class.md
+++ b/proposals/custom-states-and-state-pseudo-class.md
@@ -127,7 +127,7 @@ This feature must not have security and privacy implications.  The feature works
 
 ## Major open issues
 
--   Web components should be able to expose a list of supported states to web components users.
+-   None
 
 ## Discussions
 
--- a/proposals/custom-states-and-state-pseudo-class.md
+++ b/proposals/custom-states-and-state-pseudo-class.md
@@ -120,3 +120,91 @@ This causes compatibility issues if we ever want to introduce new CSS pseudo-cla
 ### Different syntax choices for custom states, e.g. `:--foo`
     
 We are open to other suggestions. However the discussion so far seems to favor `:state(foo)`.
+
+## Security and Privacy Considerations
+
+This feature must not have security and privacy implications.  The feature works within a single document, and nothing is exposed to outside of the document.
+
+## Major open issues
+
+-   Web components should be able to expose a list of supported states to web components users.
+
+## Discussions
+
+-   https://github.com/w3c/webcomponents/issues/738
+-   [TPAC 2019](https://www.w3.org/2019/09/17-components-minutes.html#item01)
+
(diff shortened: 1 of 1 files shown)


tkent-google:
@domenic @rakina would you review this please?

proposals/custom-states-and-state-pseudo-class.md
```diff
@@ -120,3 +120,91 @@ This causes compatibility issues if we ever want to introduce new CSS pseudo-cla
...
+
+No differences.  This feature should work regardless of first-party or third-party contexts.
+
+### 2.14. How does this specification work in the context of a user agent’s Private Browsing or "incognito" mode?
+
+No differences.  This features should work in such mode, and sites can't detect such mode with this feature.
+
+### 2.15. Does this specification have a "Security Considerations" and "Privacy Considerations" section?
+
+Yes.
```
annevk, line 202: Nothing relevant in there for this feature though, right?

proposals/custom-states-and-state-pseudo-class.md
```diff
@@ -120,3 +120,91 @@ This causes compatibility issues if we ever want to introduce new CSS pseudo-cla
...
     
 We are open to other suggestions. However the discussion so far seems to favor `:state(foo)`.
+
+## Security and Privacy Considerations
+
+This feature must not have security and privacy implications.  The feature works within a single document, and nothing is exposed to outside of the document.
+
+## Major open issues
+
+-   Web components should be able to expose a list of supported states to web components users.
```
rakina: I don't think this is a major issue, IIRC something like this was discussed in TPAC and we don't really need to provide a way to solve this

proposals/custom-states-and-state-pseudo-class.md
tkent-google: Right.
I think we need to provide "Self-Review Questionnare on Security and Privacy" anyway according to the TAG review request template.

proposals/custom-states-and-state-pseudo-class.md
tkent-google: Ok, removed it.

proposals/custom-states-and-state-pseudo-class.md
annevk: Yeah, I guess I'd lean towards answering no here. (Or "Yes, but no need to cover this feature.")

tkent-google: I think https://www.w3.org/TR/security-privacy-questionnaire/#considerations recommends to add the section even if the feature has no implication.  So this PR adds the "Security and Privacy Considerations" section.

domenic (approved):
Sorry that I missed this.

Issue #762, opened 2018-08-22
Mechanism for setting the tabindex focus flag without sprouting tabindex attribute?


alice:
[Certain HTML elements](https://html.spec.whatwg.org/multipage/interaction.html#the-tabindex-attribute:the-a-element) are focusable without having an explicit `tabindex` attribute:
* `a` elements that have an `href` attribute
* `link` elements that have an `href` attribute
* `button` elements
* `input` elements whose type attribute are not in the Hidden state
* `select` elements
* `textarea` elements
* `summary` elements that are the first summary element child of a details element

It would be nice if custom elements could also opt in to being focusable without having to sprout a `tabindex` attribute - such that if page authors removed the `tabindex` attribute, the element was still focusable.

It seems like this would be something that might fit in with whatever new API comes out of #758.

domenic:
Strong +1.

Do you think it's worth exposing both of tabindex's capabilities here, or just one? They are:

- tabindex >=0 : reachable by sequential focus navigation, and also focusable in other ways (e.g. by the mouse or by .focus())
- tabindex = -1: generally not reachable by sequential focus navigation, but focusable in other ways. (May also be reachable by sequential focus navigation in some circumstances; see the spec's "One valid reason to ignore the requirement"...)

I think there are common built-in elements whose behavior falls into each of these categories, although it may be OS dependent.

bkardell:
+1 and I would like to see both 😀

alice:
In all honesty, I've never understood why `tabindex = -1` is a thing, other than for excluding things from the tab order - i.e. why aren't all elements programmatically focusable? What do we gain from `focus()` sometimes being a no-op? Who is out here trying to call `focus()` on a `<div>` and hoping it ... fails?

But that aside, I guess allowing focusable things to also opt out of the tab order is probably useful for managing focus within complex components.

annevk:
@alice `tabindex="-1"` also affects what you can click and will get a focus ring, iirc.

alice:
@annevk True, although I've seen that cause more undesirable behaviour than desirable.

rniwa:
Whatever mechanism we use for `tabindex` should be flexible enough to default ARIA role & other ARIA attributes as well.

alice:
@rniwa Agreed, I'm hoping all of the use cases mentioned in #758, plus this one, can share the same mechanism.

SiTaggart:
[...]

PR #852, opened 2019-11-07
default focus: Update for the "hint" approach

tkent-google:


tkent-google:
@rakina @domenic @alice

proposals/custom-element-default-focus.md
```diff
@@ -36,13 +37,24 @@ class MyButton extends HTMLElement {
 
 ## Proposal
 
-Add a new property `matchFocusBehaviorOf` to `ElementInternals`, which accepts a string containing a tagname and optionally an attribute with a value (so it will accept values like `div`, `a[href]`, `input[type='checkbox']` , but not `.class`, `*`, etc.). The custom element will then be treated as if it has that tagname and attribute in terms of focus.
+Add `focusBehavior` property to `ElementInternals`, which accepts one of three string values; `"none"`, `"focusable"`, and `"infer"`.
+`"none"` indicates the element is not focusable, and `"focusable"` indicates the element's focus behaivor is same as `<div tabindex=non-negative-value>`. `"infer"` indicates the element is focusable, however a user-agent decides click behaivor and TAB behavior with hints from the element.
```
alice, line 41: So we are excluding programmatically focusable from the list of options?

tkent-google: I thought programatically-focusable was always true for `"focusable"` and `"infer"`.  The last sentence should mention programatically-focusable if there is a built-in element which is not programaticaly-focusable but is click-focusable or tab-focusable.

alice: Apologies, I should have said programmatically-focusable *only*.

tkent-google: > programmatically-focusable only.

Ah, I understand your concern.  [The last list in your comment](https://github.com/w3c/webcomponents/issues/762#issuecomment-542509589) includes "programatically focusable".

I dropped it because I was not sure if there are use-cases of the explicit programatically-focusable-only, and I thought following an existing built-in element with programatically-focusable-only by `"infer"` would be enough.

Anyway, I'd like to update [the issue thread](https://github.com/w3c/webcomponents/issues/762) with this information.

annevk:
https://github.com/w3c/webcomponents/issues/762#issuecomment-545177939 was leaning towards something like infer not being needed I thought. Was there another discussion elsewhere?

domenic:
Hmm, this is pretty confusing to me. I thought at the face to face we got consensus on the high-level concepts approach.

tkent-google:
Hmm, maybe I misunderstand the TPAC consensus.  I'm closing this PR, and will ask on the issue thread.

PR #858, opened 2019-12-05
Minor additions to CSS modules explainer based on TAG comments

dandclark:
- Add comment explaining that failed CSS module `imports` fail the whole module graph.
- Add `@import` alternative 2a where a failed `@import` doesn't have to fail the graph.


travisleithead (approved):
LGTM

PR #860, opened 2019-12-16
Strip down custom elements and SD specs

domenic:
This makes the custom elements spec redirect to the HTML Standard, and removes all contents from the shadow DOM spec that has been upstreamed, while making it clearer that what remains is just the not-yet-fully-integrated parts.

It also attempts to reduce the red squigglies in the remaining parts of the shadow DOM spec, to the best of my ReSpec knowledge.


annevk (approved):
Thanks for doing this. Happy to see this merged. I'm not entirely sure the remainder of the Shadow DOM specification is solid, but it's worth preserving the notes until it's been properly followed up on.

Issue #716, opened 2017-12-08
Scoped Custom Element Registries


justinfagnani:
Since #488 is closed, I thought I'd open up a new issue to discuss a relatively specific proposal I have for Scoped Custom Element Registries.

# Scoped Custom Element Definitions

## Overview

Scoped Custom Element definitions is an oft-requested feature of Web Components. The global registry is a possible source of name collisions that may arise from coincidence, or from **an app trying to define multiple versions of the same element**, or from more advanced scenarios like registering mocks during tests, or a component explicitly replacing an element definition for its scope.

Since the key DOM creation APIs are global, scoping definitions is tricky because we'd need a mechanism to determine which scope to use. But if we offer scoped versions of these APIs the problem is tractable. This requires that DOM creation code is upgraded to use the new scoped APIs, something that hopefully could be done in template libraries and frameworks.

This proposal adds the ability to construct `CustomElementRegistry`s and chain them in order to inherit custom element definitions. It uses `ShadowRoot` as a scope for definitions. `ShadowRoot` can be associated with a `CustomElementRegistry` when created and gains element creation methods, like `createElement`. When new elements are created within a `ShadowRoot`, that `ShadowRoot`'s registry is used to Custom Element upgrades.

### `CustomElementRegistry`

  * `CustomElementRegistry(parent?: CustomElementRegistry)`
 
    `CustomElementRegistry` is constructible, and able to inherit from a parent registry.
    
    New definitions added to a registry are not visible to the parent, and mask any registrations with the same name defined in the parent so that definitions can be overridden.

  * `CustomElementRegistry.prototype.get(name: string)`

    `get()` now returns the closest constructor defined for a tag name in a chain of registries.

  * `CustomElementRegistry.prototype.getRegistry(name: string)`

    Returns the closest registry in which a tag name is defined.

### `ShadowRoot`

`ShadowRoot`s are already the scoping boundary for DOM and CSS, so it's natural to be the scope for custom elements. `ShadowRoot` needs a `CustomElementRegistry` and the DOM creation APIs that current exist on document.

  * `customElements: CustomElementRegistry`
  
    The `CustomElementRegistry` the `ShadowRoot` uses, set on `attachShadowRoot()`.
  * `createElement()`, `createElementNS()`
    These methods create new elements using the `CustomElementRegistry` of the `ShadowRoot`.
  * `importNode()`
[...]

PR #865, opened 2020-02-21
Add Scoped CustomElementRegsitry explainer

justinfagnani:
See issue #716 

I've tried to incorporate feedback from the issue, last F2F, and reviews from colleagues.

### DIFF
--- a/proposals/Scoped-Custom-Element-Registries.md
+++ b/proposals/Scoped-Custom-Element-Registries.md
@@ -8,6 +8,8 @@ This proposal allows for multiple custom element definitions for a single tag na
 
 This is accomplished by allowing user code to create multiple custom element registries and associate them with shadow roots that function as scopes for element creation and custom element definitions. Potentially custom elements created within a scope use the registry for that scope to perform upgrades. New element construction APIs are added to ShadowRoot to allow element creation to be associated with a scope.
 
+This proposal is focused on the MVP to provide encapsulation for element definitions, and it could be extended in the future if needed to provide more versatility.
+
 ### Why do developers need scoped custom element registries?
 
 It's quite common for web applications to contain libraries from multiple sources, whether from different teams, vendors, package managers, etc. These libraries must currently contend for the global shared resource that is the `CustomElementRegistry`. If more than one library (or more than one instance of a library) tries to define the same tag name, the application will fail. 
@@ -36,9 +38,11 @@ This proposal allows user code to create new instances of `CustomElementRegistry
 
 ```js
 const registry = new CustomElementRegistry();
+```
 
 and associate them with a ShadowRoot:
 
+```js
 export class MyElement extends HTMLElement {
   constructor() {
     this.attachShadow({mode: 'open', registry});
@@ -64,9 +68,9 @@ this.shadowRoot.innerHTML = `<other-element></other-element>`;
 
 These scoped registries will allow for different parts of a page to contain definitions for the same tag name.
 
-### Creating and using a CustomElementRegistry
+### Creating and using a `CustomElementRegistry`
 
-A new CustomElementRegistry is created with the CustomElementRegistry constructor, and attached to a ShadowRoot with the `registry` option to `HTMLElement.prototype.attachShadow`:
+A new `CustomElementRegistry` is created with the `CustomElementRegistry` constructor, and attached to a ShadowRoot with the `registry` option to `HTMLElement.prototype.attachShadow`:
 
 ```js
 import {OtherElement} from './my-element.js';
@@ -91,36 +95,13 @@ In order to support scoped registries we add new scoped APIs, that were previous
 * `createElementNS()`
 * `importNode()`
 
-These APIs work the same as their Document equivalents, but use scoped registries instead of the global registry.
-
-### Registry Inheritance
-
-In a shadow root with a scoped registry, all element creation APIs use that shadow root's registry, and not the document's, to look up definitions. To inherit definitions from the global, or another scoped registry, a parent registry can be passed in at construction time:
-
-```js
-const registry = new CustomElementRegistry({
-    parent: window.customElements,
-});
-```
-
-This inheritance is live. New registrations added to the parent registry are available to inheriting registries. This is useful in the case where an element is already written to use the global registry, but needs to register a private helper element, or override only a single element in conflict with the global. It may also be useful in plug-in architectures where the host program provides a number of elements to plugins.
-
-For non-live inheritance, we can add a method to CustomElementRegistry that returns all of its registrations:
-
-```js
-const registry = new CustomElementRegistry({
-    definitions: {
-      ...window.customElements.getDefinitions(),
-      'local-element': LocalElement,
-    }
-});
-```
+These APIs work the same as their `Document` equivalents, but use scoped registries instead of the global registry.
 
 ### Finding a custom element definition
 
 Because there is no longer a single global custom element registry, when creating elements, the steps to look up a custom element definition need to be updated to be able to find the correct registry.
 
-That process needs to take a context node that is used to look up the definition. The registry is found by getting the context node's root. If the root has a CustomElementRegistry, use that registry to look up the definition, otherwise use the global objects CustomElementRegistry object.
+That process needs to take a context node that is used to look up the definition. The registry is found by getting the context node's root. If the root has a `CustomElementRegistry`, use that registry to look up the definition, otherwise use the global objects CustomElementRegistry object.
 
 The context node is the node that hosts the element creation API that was invoked, such as `ShadowRoot.prototype.createElement()`, or `HTMLElement.prototype.innerHTML`. For `ShadowRoot.prototype.createElement()`, the context node and root are the same.
 
@@ -134,49 +115,9 @@ Another option for looking up registries is to store an element's originating re
 
 Constructors need special care with scoped registries. With a single global registry there is a strict 1-to-1 relationship between tag names and constructors. Scoped registries change this by allowing the same tag name to be associated with multiple constructors, which is solved by the altered look up a custom element definition process allowing the browser to find the correct constructor given a tag name.
 
-In the other direction, we want to be able to call `new MyElement()`, which means we need to be able to locate the correct tag name from a constructor as well.
+As a result, it must limit constructors by default to only looking up registrations from the global registry. If the constructor is not defined in the global registry, it will throw.
 
-The way this is done is by limiting constructors by default to only looking up registrations from the global registry. If the constructor is not defined in the global registry, it will throw. In order to get a constructor that creates a scoped definition, customElementRegistry.define() returns a new constructor:
-
-```js
-import {OtherElement} from './my-element.js';
-
-const registry = new CustomElementRegistry();
-
-// define() returns a new class:
-const LocalOtherElement = registry.define('other-element-2', OtherElement);
-const el = new LocalOtherElement();
-el.tagName === 'other-element-2';
-
-// The same class is available from registry.get():
-const O = registry.get('other-element-2')
-const el2 = new O();
-el2.tagName === 'other-element-2';
-```
-
-The constructor returned by `define()` is from a trivial subclass of the registered class.
-
-### Sugar for bulk registrations
-
-To make importing and registering multiple definitions easier, the CustomElementRegistry constructor can take an object containing multiple definitions:
-
-```js
-import {ElementOne} from './element-one.js';
-import {ElementTwo} from './element-two.js';
-
-const registry = new CustomElementRegistry({
-  definitions: {
-    'element-one': ElementOne,
-    'element-two': ElementTwo,
-  }
-});
-
-export class MyElement extends HTMLElement {
-  constructor() {
-    this.attachShadow({mode: 'open', registry});
-  }
-}
-```
+This poses a limitation for authors trying to use the constructor to create new elements associated to scoped registries but not registered as global. More flexibility can be analyzed post MVP, for now, a user-land abstraction can help by keeping track of the constructor and its respective registry.
 
 ## API
 
@@ -184,17 +125,13 @@ export class MyElement extends HTMLElement {
 
 #### Constructor
 
-The CustomElementRegistry constructor creates a new instance of CustomElementRegistry, independent of the instance available at `window.customElements`.
+The `CustomElementRegistry` constructor creates a new instance of CustomElementRegistry, independent of the instance available at `window.customElements`.
 
 ##### Syntax
 
 ```js
-const registry = new CustomElementRegistry({parent, definitions});
+const registry = new CustomElementRegistry();
 ```
-
-##### Parameters
-parent A parent CustomElementRegistry to inherit definitions from.
-definitions An object whose keys are custom element names and values are the associated constructors.
 
 ### ShadowRoot
 
@@ -205,6 +142,24 @@ ShadowRoot adds element creation APIs that were previously only available on Doc
 * `importNode()`
 
 These are added to provide a root and possible registry to look up a custom element definition.
+
+## Open Questions
+
+### Adopted elements and Scoped Registry
+
+There are concern about what happens when an element with a custom registry moves to another document and the GC implications. We debated briefly about possible solutions:
+
+1. make the registry a simple key, value pairs that can be moved around.
+2. recreate a new registry entry when moving the element, and let the author to repopulate it in `adoptedCallback`. This is problematic because the registry is created before attaching the shadowRoot.
+3. create a new callback that can receive the new registry when moved. This is problematic as well because what happen when the registries are coming from another library, already created by someone else?
+4. find ways for implementers to preserve the original registry (ideal).
+
+### Intersection with Declarative Shadow Root
+
+Although these two proposals are in early stages, we need to solve the intersection semantics. There are two main issues:
+
+1. if a declarative shadow root is created, elements inside that shadow should not be upgraded with the global registry. a possible solution is to add a new attribute, similar to `mode` to indicate to the parser that a custom registry is going to be eventually associated to this shadow.
+2. if the component is planning to reuse the instance of the declarative shadow root (which is ideal), how can the component associate that instance with a registry? this indicates that maybe the association between ShadowRoot and CustomElementRegistry cannot be defined via `attachShadow()`, and instead, something like `ElementInternals` is much more flexible.
 
 ##  Alternatives to allowing multiple definitions
 
--- a/proposals/Scoped-Custom-Element-Registries.md
+++ b/proposals/Scoped-Custom-Element-Registries.md
@@ -1,0 +1,227 @@
+# Scoped Custom Element Registries
+
+Author: [Justin Fagnani](https://github.com/justinfagnani)
+
+## Introduction
+
+This proposal allows for multiple custom element definitions for a single tag name to exist within a page. 
+
+This is accomplished by allowing user code to create multiple custom element registries and associate them with shadow roots that function as scopes for element creation and custom element definitions. Potentially custom elements created within a scope use the registry for that scope to perform upgrades. New element construction APIs are added to ShadowRoot to allow element creation to be associated with a scope.
+
+### Why do developers need scoped custom element registries?
+
+It's quite common for web applications to contain libraries from multiple sources, whether from different teams, vendors, package managers, etc. These libraries must currently contend for the global shared resource that is the `CustomElementRegistry`. If more than one library (or more than one instance of a library) tries to define the same tag name, the application will fail. 
+
+Multiple versions of a library is a common source of this problem. This can happen for many reasons, but there are a few application/library types where this is common:
+
+1. **Any sized applications using npm:** NPM may install multiple versions of a library if it can't find a single version that satisfies version constraints (and it many cases where it could, but doesn't because of its lack of a version constraint solver). Thus an application may contain multiple definitions of a tag name from the same library, but at different versions.
+2. **Large, complex applications:** These can have components authored by different teams, with loose coordination between them, built and deployed to production at different times. This makes library duplication in production common.
+3. **Applications with plug-in architectures:** Some applications that allow plug-ins will run plug-ins in the main window. The application and plug-ins will typically be built and distributed independently. If the application and plug-in use the same component, they will have contention for the tag name.
+4. **Browser Extensions:** Many browser extensions create DOM to be displayed in the main page. If this DOM is created with custom elements, they will need to be registered and could conflict with the page's scripts.
+
+In addition to library duplication, there are other common use-cases for scopes:
+
+1. **Third-party, CDN-hosted libraries:** Libraries like social-media share buttons and embeds, maps and documentation viewers, etc., can vend complex UI widgets that may need to register many elements, and they should not be using up a global namespace that they don't fully control.
+2. **Internal implementation elements:** A component may have private internal elements that it needs to register, but would prefer not to leak to the external environment.
+
+### Why use shadow roots as registration scopes?
+
+Shadow roots provide an encapsulation mechanism for isolating a DOM tree's, nodes, styles, and events from other scopes. The goal is to allow scopes to function without required coordination with other scopes. The globally shared custom element registry, however, requires coordination so that multiple scopes on a page agree on the registrations. Rather than invent a new scoping primitive to the DOM, it is natural to extend the shadow root scoping responsibilities to include custom element registrations.
+
+This approach also allows for a nice API by extending the DocumentOrShadowRoot interface and `ShadowRoot#attachShadow()`.
+
+## Overview
+
+This proposal allows user code to create new instances of `CustomElementRegistry`:
+
+```js
+const registry = new CustomElementRegistry();
+
+and associate them with a ShadowRoot:
+
+export class MyElement extends HTMLElement {
+  constructor() {
+    this.attachShadow({mode: 'open', registry});
+  }
+}
+```
+
+Such scoped registries can be populated with element definitions, completely under the control of the registry owner:
+
+```js
+import {OtherElement} from './my-element.js';
+
+registry.define('other-element', OtherElement);
+```
+
+Definitions in this registry do not apply to the main document, and vice-versa. The registry must contain definitions for all elements used.
+
+Once a registry and scope are created, element creation associated with the scope will use that registry to look up custom element definitions:
+
+```js
+this.shadowRoot.innerHTML = `<other-element></other-element>`;
+```
+
+These scoped registries will allow for different parts of a page to contain definitions for the same tag name.
+
+### Creating and using a CustomElementRegistry
+
+A new CustomElementRegistry is created with the CustomElementRegistry constructor, and attached to a ShadowRoot with the `registry` option to `HTMLElement.prototype.attachShadow`:
+
+```js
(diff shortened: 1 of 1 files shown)


justinfagnani:
Reviewing the issue thread again, it may be good to add some non-goals, like that this isn't a security primitive. Or should that be a well-known non-goal of Shadow DOM in general?

proposals/Scoped-Custom-Element-Registries.md
```diff
@@ -0,0 +1,227 @@
...
+This approach also allows for a nice API by extending the DocumentOrShadowRoot interface and `ShadowRoot#attachShadow()`.
+
+## Overview
+
+This proposal allows user code to create new instances of `CustomElementRegistry`:
+
+```js
+const registry = new CustomElementRegistry();
+
+and associate them with a ShadowRoot:
```
LarsDenBakker, line 43:
`````suggestion
```
and associate them with a ShadowRoot:
```js
`````

**Comment by justinfagnani** [comment]
cc @rniwa @mfreed7 @annevk

**Comment by LarsDenBakker** [comment]
In previous drafts registries were inherited through the DOM hierarchy. Is this intentionally out of the current proposal? I think it's a good thing, but just making sure.

If an element is in one shadow root and moves to another shadow root, it's element definition is still the one it received from the initial shadow root, correct? That wasn't fully clear to me from the wording.

For the list in `Why do developers need scoped custom element registries?` I think there are two more benefits:
- Scoped registries allow creating stronger associations between a web component and it's dependencies. Right now we need to rely on side effects on the global registry. It's easy to forget importing a component but not notice it because another file has already imported it.
- Scoped registries allow creating tree-shakable component libraries because they can export only classes. Right now they need to expose many entrypoints, one for each side effect.

**Comment by justinfagnani** [comment]
@LarsDenBakker 

> In previous drafts registries were inherited through the DOM hierarchy. Is this intentionally out of the current proposal? I think it's a good thing, but just making sure.

It's very much intentional and based on discussion and feedback. The goal is to allow a component to safely operate without interference from other scopes. If a component uses a scoped registry, we presume it doesn't want interference from outer scopes. If it does, then it has to opt into inheritance.

> If an element is in one shadow root and moves to another shadow root, it's element definition is still the one it received from the initial shadow root, correct? That wasn't fully clear to me from the wording.

No, it will use the registry from the shadow root it's _currently_ in. Moving elements between scopes is very, very rare.

> For the list in Why do developers need scoped custom element registries? I think there are two more benefits:

Thanks for those! I'll add them.

**Comment by LarsDenBakker** [comment]
> No, it will use the registry from the shadow root it's currently in. Moving elements between scopes is very, very rare.

I mean a situation like this:

```js
import { ElementA } from './element-a.js';
import { ElementB } from './element-b.js';

const a = document.body.appendChild(document.createElement('div'));
const b = document.body.appendChild(document.createElement('div'));

a.attachShadow({
  mode: 'open',
  registry: new CustomElementRegistry({
    definitions: { 'my-element': ElementA },
  }),
});
b.attachShadow({
  mode: 'open',
  registry: new CustomElementRegistry({
    definitions: { 'my-element': ElementB },
  }),
});

a.shadowRoot.innerHTML = '<my-element></my-element>';
const myElement = a.shadowRoot.querySelector('my-element');

console.assert(myElement.constructor === elementA);

b.shadowRoot.appendChild(myElement);
console.assert(myElement.constructor === elementA);
```

So even after moving shadow roots, the constructor stays the same.

bicknellr:
> So even after moving shadow roots, the constructor stays the same.

Yes, I don't think there's any desire here to try to swap definitions somehow once they've been created.

proposals/Scoped-Custom-Element-Registries.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Scoped element creation APIs
+
+Element creation APIs, like `createElement()` and `innerHTML` can be grouped into global API (those on Document or Window) and scoped APIs (those on HTMLElement and ShadowRoot). The scoped APIs have an associated Node that can be used to look up a CustomElementRegistry and thus a scoped definition.
+
+In order to support scoped registries we add new scoped APIs, that were previously only available on `Document`, to `ShadowRoot`:
+
+* `createElement()`
+* `createElementNS()`
+* `importNode()`
```
caridy, line 96: what about `adoptNode`?

proposals/Scoped-Custom-Element-Registries.md
```diff
@@ -0,0 +1,227 @@
...
+
+Because there is no longer a single global custom element registry, when creating elements, the steps to look up a custom element definition need to be updated to be able to find the correct registry.
+
+That process needs to take a context node that is used to look up the definition. The registry is found by getting the context node's root. If the root has a CustomElementRegistry, use that registry to look up the definition, otherwise use the global objects CustomElementRegistry object.
+
+The context node is the node that hosts the element creation API that was invoked, such as `ShadowRoot.prototype.createElement()`, or `HTMLElement.prototype.innerHTML`. For `ShadowRoot.prototype.createElement()`, the context node and root are the same.
+
+#### Note on looking up registries
+
+One consequence of looking up a registry from the root at element creation time is that different registries could be used over time for some APIs like HTMLElement.prototype.innerHTML, if the context node moves between shadow roots. This should be exceedingly rare though.
```
caridy, line 110: this sentence was a little confusing at first, maybe need some rewording!

proposals/Scoped-Custom-Element-Registries.md
justinfagnani: `adoptNode` doesn't create nodes or run upgrade steps (https://dom.spec.whatwg.org/#concept-node-adopt) so it doesn't need to find a registry.

proposals/Scoped-Custom-Element-Registries.md
justinfagnani: I'll try to clean this up.

bicknellr:
I don't think inheritance is necessary to get basically all of the benefits and might be worth leaving out of first iteration of this. It opens the door for accidental dependence on transitive dependencies (for example, component set authors creating / exporting a registry with their package's components in one big bundle) and adds complexity to the lookup process that makes it harder to understand where your definitions come from, when you should generally know where the elements you depend on are defined.

Separate topic: I'd like to recommend that `.getDefinitions()` and the `definitions` option in the constructor object use Map-like entries since registries are basically a fancy map and this is how Map iterators / `.entries()` work. (`{definitions: {'x-a': A, 'x-b': B}}` would become `{definitions: [['x-a', A], ['x-b', B]]}`)

You can still use object literals for sugar by wrapping them in `Object.entries`. For example: `{definitions: Object.entries({'x-a': A, 'x-b': B})}`.

caridy:
Feedback after debating this update with @diervo and @JanMiksovsky today:

1. Basic API looks good.
2. Inheritance is a convenience, otherwise complex apps will have to create big and probably expensive registries for the different buckets of components. (we can survive without it).
3. `getDefinitions` is a new capability that seems problematic, and unnecessary. up until now the web have survived without providing the list of built-in tags and their constructor. doing so for custom elements only seems weird, but we are more worry about someone having a way to inspect all defined elements, seems like a big footgun. We recommend remove that from the proposal for now.
4. `definitions` option is a convenience, but easy to do in user-land via subclassing. recommendation is to remove it from the MVP.
5. about "Note on looking up registries", we agreed that a dynamic lookup is sufficient.
6. about "Custom element constructors", we need to think about it, we agreed that only looking up on the global registry seems like a viable solution, and restrictive enough, but @JanMiksovsky want to think more about it (we will keep you posted).

LarsDenBakker:
It would be great if we could merge the current agreed status. Is there something I can help with?

leobalter:
@LarsDenBakker thanks for the reminder.

For now we are tracking the progress in a separate repo (for multiple specific issues) in here: https://github.com/justinfagnani/scoped-custom-elements

Should we update this PR with the new content or perhaps just a link to the new repo?

I might be wrong, but the current status is:

- We agreed to follow up the items previously summarized by @caridy in this [comment above](https://github.com/w3c/webcomponents/pull/865#issuecomment-592863359).
- We will also work on a separate proposal for [a bulk define method for components](https://github.com/justinfagnani/scoped-custom-elements/issues/2).
- I am - slowly - [preparing the work for a TAG review](https://github.com/justinfagnani/scoped-custom-elements/issues/4), I'm trying not to rush on drafting the specs as it's challenging for a new comer (myself).
  - Ideally I'll have the html draft ready
  - and I should follow up with WPT tests ready
  - Work on [getting feedback from Mozilla](https://github.com/mozilla/standards-positions/issues/424).

cc'ing @justinfagnani @caridy here.

what's the best approach for this PR now?

LarsDenBakker:
Thanks for the update. Besides the PR it would also be good to update the issue description here https://github.com/w3c/webcomponents/issues/716 to either reflect the new proposal or link to the new repository. 

I've had a few people get confused about what exactly the current status of the proposal is.

leobalter:
Sure thing! For that I need to coordinate w/ @justinfagnani and ask him to update the issue description. I don't have write access to change issues in this w3c/webcomponents repo so it's a tricky job for a team. :)

leobalter:
after @rniwa's feedback, we are moving things back to this repo.

In this case, let's update this PR with the latest explainer from https://github.com/justinfagnani/scoped-custom-elements.

justinfagnani:
@leobalter @caridy @LarsDenBakker  I just updated the PR with the changes that landed in https://github.com/justinfagnani/scoped-custom-elements

leobalter:
Thanks!

cc @rniwa. I hope this proposal is now a bit closer to fair clarification and up to a point we can work through any technical concerns. I'm gonna try to summarize the current topics after I recompose myself from the TC39 meeting this week, please let me know if there is any other outstanding parts to look at.

rniwa:
The latest proposal still doesn't answer my question at https://github.com/w3c/webcomponents/issues/716#issuecomment-679426920

justinfagnani:
@rniwa I thought we agreed in the last virtual f2f that shadow roots with scoped registries should not inherit from the main registry - once you add a scoped registry you need to add definitions for all elements you use.

So the explainer has this line in the Overview

> Definitions in this registry do not apply to the main document, and vice-versa. The registry must contain definitions for all elements used.

and is also implied in the * Finding a custom element definition* section:

> That process needs to take a context node that is used to look up the definition. The registry is found by getting the context node's root. If the root has a CustomElementRegistry, use that registry to look up the definition, otherwise use the global objects CustomElementRegistry object.

That might not standout enough though, do you have a suggestion for how to make that more clear? I don't have a section on upgrades because it seems like upgrading is untouched by this change, only the "find a registry" steps are and everything past that is the same.

Maybe that's not quite right though and for upgrades the tree walk to find elements to upgrade needs to be modified too?

leobalter:
> The latest proposal still doesn't answer my question at [#716 (comment)]
(https://github.com/w3c/webcomponents/issues/716#issuecomment-679426920)

@rniwa ofc! by no means I tried to get away with this. My goal was to make sure we have the discussions in this repo. We still have that original issue and this current PR, but none are resolved yet.

To sum up with @justinfagnani's comment, I wonder if @caridy's comment at https://github.com/w3c/webcomponents/issues/716#issuecomment-681158124 might help with some potential loose ends.

I hope we can work through those parts, and I'm sorry if I gave any impression of not being fair with your concerns. I appreciate the process where we move this forward in collaboration and shared efforts.

rniwa:
I'm talking about the upgrade ordering. It's unclear in what order custom elements get upgraded when multiple shadow references to a give scope.

justinfagnani:
Ok, so I think we're looking for clarification and changes to steps 17 and 18 in the Element definition_ steps: https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition

It seems like these steps need to be branched depending on whether the registry is the global registry or a scoped registry.

If it's the global registry:

- 17 stays the same:

   > Let document be this CustomElementRegistry's relevant global object's associated Document.

- 18 needs to be updated to exclude shadow roots with scoped registries:
  >  Let upgrade candidates be all elements that are shadow-including descendants of document, whose namespace is the HTML namespace and whose local name is localName, in shadow-including tree order. Additionally, if extends is non-null, only include elements whose is value is equal to name. _Exclude all elements whose root is a shadow root has a scoped custom element registry._

  or something like that.

If it's a scoped registry:

- Step 17 might need to get all the shadow roots associated with the registry. We can define an order for connected shadow roots by using either tree-order or the order in which the shadow roots were created.

   Creation order might be more efficient if registries have to keep a list of associated roots anyway. In that case we need to clarify if disconnected shadow roots are upgraded.

  If keeping references from registries to shadow roots is not desired (can they be weak references?), then we would need get the associated roots from the document. The order would then be document order and naturally exclude disconnected shadow roots.

- Step 18 starts with candidates that are shadow-excluding descendants of the roots from the previous step

I think in the explainer it would make sense to have a new section before *Finding a custom element definition* like *Element definition* to describe this.

@rniwa does this sound about right?

@rniwa @mfreed7 do you have guidance on whether registries should keep references to the associated shadow roots?

caridy:
> I'm talking about the upgrade ordering. It's unclear in what order custom elements get upgraded when multiple shadow references to a give scope.

@rniwa I'm still unclear on how is this different from 1 registry (the global one) vs 10 registries? can you elaborate more?

> @rniwa @mfreed7 do you have guidance on whether registries should keep references to the associated shadow roots?

This should not be different to what we do today with the global registry. New entries on those registries can cause upgrades on multiple shadow roots, isn't it? Just like adding a new entry into a scoped registry can cause similar upgrades, in the same fashion, including same ordering.

justinfagnani:
@caridy 

> I'm still unclear on how is this different from 1 registry (the global one) vs 10 registries? can you elaborate more?

The current spec has a definitive order for upgrades. With a scoped registry associated with multiple shadow roots, there are at least two plausible orderings for which shadow roots, and thus which elements, to upgrade first: shadow-including document order, or shadow root creation order (order that shadow roots were associated with the registry).

> This should not be different to what we do today with the global registry. New entries on those registries can cause upgrades on multiple shadow roots, isn't it? Just like adding a new entry into a scoped registry can cause similar upgrades, in the same fashion, including same ordering.

We _may_ want to pick the option where we use shadow root creation order because we wouldn't have to walk the entire shadow-including document, we could just visit each associated shadow root and consider only descendants. That seems like a faster choice, though there could be an inconsistency with the global registry if we include disconnected shadow roots.

Keeping a list of shadow roots with a registry would create a cycle between shadow roots and registries. I don't know if that's something generally avoided in the DOM, which is why I'm asking for the guidance.

rniwa:
> @caridy
> 
> > I'm still unclear on how is this different from 1 registry (the global one) vs 10 registries? can you elaborate more?
> 
> The current spec has a definitive order for upgrades. With a scoped registry associated with multiple shadow roots, there are at least two plausible orderings for which shadow roots, and thus which elements, to upgrade first: shadow-including document order, or shadow root creation order (order that shadow roots were associated with the registry).

Shadow-including document order is still ambiguous. Some shadow trees may not be connected to a document, and there could be multiple top-level documents in different windows.

> > This should not be different to what we do today with the global registry. New entries on those registries can cause upgrades on multiple shadow roots, isn't it? Just like adding a new entry into a scoped registry can cause similar upgrades, in the same fashion, including same ordering.
> 
> We _may_ want to pick the option where we use shadow root creation order because we wouldn't have to walk the entire shadow-including document, we could just visit each associated shadow root and consider only descendants. That seems like a faster choice, though there could be an inconsistency with the global registry if we include disconnected shadow roots.

There will definitely be inconsistencies here.

> Keeping a list of shadow roots with a registry would create a cycle between shadow roots and registries. I don't know if that's something generally avoided in the DOM, which is why I'm asking for the guidance.

So creation order is problematic if we also upgrade elements in disconnected shadow trees because then we end up having to keep all shadow trees that uses the same custom element registry alive (i.e. prevent GC) if at least one of the shadow root has to be kept alive because whether a custom element will be upgraded or not is script observable. This is the exact reason we had to make upgrades only work with connected nodes.

justinfagnani:
I don't know what kind of reference facilities are available to use in the spec. Can we have have registries maintain essentially an ordered weak enumerable set of shadow roots? Then on definition we can only upgrade the connected shadow roots.

rniwa:
> I don't know what kind of reference facilities are available to use in the spec. Can we have have registries maintain essentially an ordered weak enumerable set of shadow roots? Then on definition we can only upgrade the connected shadow roots.

Even connected shadow roots is a bit problematic because it would mean any shadow root connected to a document of a removed iframe would be included. This is actually a tough problem.

Maybe @annevk has some idea...

annevk:
It's been a while since I thought about this and while I tried to read up, I might have missed something, but if we scope a custom registry to a document and don't allow it to be used by multiple documents (i.e., make it as close to the global registry as possible), would you still have that problem? At least it seems you could determine _upgrade candidates_ in `define()` in near-identical fashion. If you want multiple documents then I think @justinfagnani is correct that we should keep track of them (and also clean them up when we stop caring about them). (It seems better to track documents than shadow roots so that the ordering within a document is the same as with the global registry.)

justinfagnani:
@annevk 
> It seems better to track documents than shadow roots so that the ordering within a document is the same as with the global registry

Naively, it seems like a potential advantage of and disadvantage of scoped registries is perf:
- A potential improvement i that skipping a shadow-including document tree walk when you have a registry associated with a single shadow root seems like a nice perf win.
- A potential downside is if pages start having a lot of registries (1 per element definition x 100s of definitions) then scoped registries could drastically increase the number of walks.

On my team we have definitely seen registration cost as a problem in very WC-heavy pages (100s of definitions) and devised ways to delay registration to improve first paint.

annevk:
So how well established are the requirements? In terms of supporting multiple shadow roots, documents, etc. And by registration cost do you mean define/upgrade cost? It seems that running JavaScript would easily be the most expensive there. (Not saying that we should accept large tree walks if we can avoid them, but consistency and simplicity in the high-level design are also worthy goals.)

rniwa:
> So how well established are the requirements? In terms of supporting multiple shadow roots, documents, etc. And by registration cost do you mean define/upgrade cost? It seems that running JavaScript would easily be the most expensive there. (Not saying that we should accept large tree walks if we can avoid them, but consistency and simplicity in the high-level design are also worthy goals.)

If moving a shadow root with a scoped registry is more of an edge case, then we could also consider not upgrading anything that moved to another document and only consider shadow roots that are connected to the same document. That would allow us to keep the algorithm simple & similar.

justinfagnani:
I think supporting multiple shadow roots is the only really important requirement, so that one registry can be shared across all instances of a definition.

Other than that, I haven't heard of requests for supporting multiple documents. Sometimes printing / screenshot libraries will clone a DOM tree into an iframe, which is something to consider, but I think the ones that handle shadow roots aren't also registering any elements in the iframe - they're just deeply cloning un-upgraded nodes and adding shadow roots as needed.

Agreed that the major cost of definitions is the JS time, but for 100s of definitions and 1000s of elements (largest app I know of is ~1k definitions, ~7k nodes) this could be non-trivial, especially if every definition has its own registry.

caridy:
I'm on board with the idea of not allowing sharing registries across documents, we can always expand on that in the future if it becomes an issue.

I'm not ok with not allowing sharing a registry between multiple shadow roots. Even though we don't plan to do so at salesforce platform ATM, I believe it will feel very natural for anyone working on a group of components, to share the same registry. Preventing that sounds very restrictive.

annevk:
The other reason I like the idea that custom registries use the same tree walk as the global registry is that implementation optimizations will benefit both.

It's a good question what should happen when you move a shadow tree across documents. And losing the registry without recourse feels somewhat sloppy. Perhaps we need the equivalent of `adoptedCallback` for shadow roots so you get to recover somehow. (Alternatively we only support this on shadow roots created on ElementInternals so you can use `adoptedCallback` to recover.)

rniwa:
> The other reason I like the idea that custom registries use the same tree walk as the global registry is that implementation optimizations will benefit both.

I'm not certain that there is a meaningful optimization we could implement for the global one. The idea of visiting shadow trees in the order they were associated with a custom element registry works because `define` or `upgrade` calls per each custom element registry and the registry can remember its insertion order. This is somewhat independent of the discussion of whether a shadow trees that got moved to another document or ones that are disconnected from either the original or another document will be considered for upgrading or not.

> It's a good question what should happen when you move a shadow tree across documents. And losing the registry without recourse feels somewhat sloppy. Perhaps we need the equivalent of `adoptedCallback` for shadow roots so you get to recover somehow. (Alternatively we only support this on shadow roots created on ElementInternals so you can use `adoptedCallback` to recover.)

I'm a bit confused here. You don't need `ElementInternals` to recover, right? We'd just expose this on `ShadowRoot` then whoever has access to `ShadowRoot` can get the registry to get the the necessary definitions, etc...

annevk:
I was thinking the global one could do shadow trees with their own registry more quickly, but it depends a bit on nested shadow trees and might require more bookkeeping than it's worth. It does seem like you want that to be fast as well though since not all elements will be nested in a shadow tree.

---

If a registry is bound to a document and you move a shadow root associated with a registry to another document, I would assume we want that shadow root to lose its connection to the registry. I suppose we could also make a bunch of things no longer work, but either way you should be able to recover from that I think (by associating the shadow root with an equivalent registry for that document). For elements we offer `adoptedCallback`, for shadow roots we don't have such a thing. So either we'd add something to shadow roots or we tightly couple these kind of shadow roots (that have a custom registry) to elements that can have custom internals.

rniwa:
> I was thinking the global one could do shadow trees with their own registry more quickly, but it depends a bit on nested shadow trees and might require more bookkeeping than it's worth. It does seem like you want that to be fast as well though since not all elements will be nested in a shadow tree.

Well, you'd need nested shadow trees to be ordered so you kind of need to either cache the ordering or traverse through them anyway. The only work you're gonna avoid for free will be leaf shadow trees or shadow trees in which there are no more than one nested shadow tree which uses the global registry. But that's easy enough of an optimizations to add regardless of what we do with scoped registries. It would be really shitty if we had to traverse through the entire document to figure out which shadow tree comes first each time upgrading step needs to run in `define` call.

> If a registry is bound to a document and you move a shadow root associated with a registry to another document, I would assume we want that shadow root to lose its connection to the registry. I suppose we could also make a bunch of things no longer work, but either way you should be able to recover from that I think (by associating the shadow root with an equivalent registry for that document). For elements we offer `adoptedCallback`, for shadow roots we don't have such a thing. So either we'd add something to shadow roots or we tightly couple these kind of shadow roots (that have a custom registry) to elements that can have custom internals.

What do you mean by "lose its connection"? Meaning that things like innerHTML will no longer work? That's sort of in line with what I was suggesting earlier but for this recovery thing to work, we'd need to make it possible to re-associate a shadow root with a new scoped registry after the fact. That seems like a new complication to me.

Overall, there is a lot of open questions with regards to upgrade ordering & multiple documents.

annevk:
Yes, it's a complication, but it also seems somewhat necessary for a complete design of this feature. I could see not supporting it if we have a viable path to add it in the next version though. Basically, we need to tackle multiple documents in one of two ways:

1. Support multiple documents, even though this is weirdly inconsistent with the global registry which only supports a single document.
2. Support developers with some kind of API if they want to handle cross-document inserts. Potentially we don't ship this initially and whenever you move a shadow root with a custom registry across document boundaries it's broken. However, we should have a concrete design to add it in the future as moving nodes across documents is fairly common and not unreasonable.

(Note that the way custom elements / the global registry offer support here is through the `adoptedCallback`, which allows you to change things around as needed.)

rniwa:
So there is another interesting piece here. If a scoped registry has any kind of fallback to the global registry, then it would kill the possibility to ever support inheritance via the tree ancestry. This is because both ancestor shadow roots' scoped registry as well as the global may define a custom element of the same name, and if we ever introduced a direct fallback from a nested shadow root to the global, it won't be backwards compatible to then introduce a fallback through shadow tree ancestors.

This feature seems to pose a lot of tricky design pitfalls... We may need to think about the whole problem space (as in everything we may want to do in the future) before we can even come up with MVP. Someone should try prototyping this in some browser and figure out how it works out in practice too. This is a very tricky API to design.

justinfagnani:
> So there is another interesting piece here. If a scoped registry has any kind of fallback to the global registry, then it would kill the possibility to ever support inheritance via the tree ancestry. This is because both ancestor shadow roots' scoped registry as well as the global may define a custom element of the same name, and if we ever introduced a direct fallback from a nested shadow root to the global, it won't be backwards compatible to then introduce a fallback through shadow tree ancestors.

I think this is one of the motivations behind not having a fallback. Once you have a scoped registry you have to fully specify all definitions. Evolving from there any inheritance or fallback would have to be opt-in.

caridy:
> > So there is another interesting piece here. If a scoped registry has any kind of fallback to the global registry, then it would kill the possibility to ever support inheritance via the tree ancestry. This is because both ancestor shadow roots' scoped registry as well as the global may define a custom element of the same name, and if we ever introduced a direct fallback from a nested shadow root to the global, it won't be backwards compatible to then introduce a fallback through shadow tree ancestors.
> 
> I think this is one of the motivations behind not having a fallback. Once you have a scoped registry you have to fully specify all definitions. Evolving from there any inheritance or fallback would have to be opt-in.

Certainly that's the case here! And we are counting on that! Honestly, I don't think we will want to explore the fallback mechanism in the future, I hope we don't, but we will see!

annevk:
@justinfagnani could you maybe update the PR to include these design points? Both the open questions around ordering and how to deal with multiple documents and why we're not handling fallback (right now). I guess we should have merged this and discussed these things in dedicated issues, but here we are.

It seems you also need to associate your account with a W3C account, although given that this repository isn't really bound to a Working Group afaik I'm not sure that matters a whole lot...

leobalter:
>  I guess we should have merged this and discussed these things in dedicated issues, but here we are.

if we get this merged I can work through these additional topics with @caridy in a quick follow up. Otherwise I might direct a patch to @justinfagnani's branch.

The goal is not to raise pressure but facilitate the process. Please let me know what would work best.

leobalter:
I opened #895 to redirect the design issues in the current specs being highlighted here. Maybe this could unblock merging this current explainer as well?

> Both the open questions around ordering and how to deal with multiple documents and why we're not handling fallback (right now). I guess we should have merged this and discussed these things in dedicated issues, but here we are.

@annevk I'm adding this bit to the explainer this week, if it does get merged before that, I'll have a direct PR here.

justinfagnani:
@annevk and @leobalter 

> @justinfagnani could you maybe update the PR to include these design points?

I can. I'm a bit swamped right now, so it'll be a few more days. Could we go with merging and opening issues like @leobalter has done with #895? Otherwise he has merge permissions on my fork. Whatever's best.

annevk:
You have to resolve the ipr check before I can merge this.

To be clear, I don't think these issues are orthogonal in any way. They have to be resolved to get agreement on this proposal.

leobalter:
Thanks for correcting me, @annevk. I was biased on my previous statement and I'll fix the wording and make sure we don't skip any steps here.

annevk:
@justinfagnani you said the IPR thing was resolved, but it still looks red from here.

justinfagnani:
@annevk I created my w3.org account and associated with my Github account. Is there anything more I need to do? I don't have permissions to revalidate the PR.

annevk (approved):

justinfagnani:
thanks @annevk !

PR #866, opened 2020-02-27
Update Imperative-Shadow-DOM-Distribution-API.md

yuzhe-han:
Updated based on the TPAC F2F meeting in Sept, 2019. https://github.com/whatwg/html/issues/3534#issuecomment-537802687

### DIFF
--- a/proposals/Imperative-Shadow-DOM-Distribution-API.md
+++ b/proposals/Imperative-Shadow-DOM-Distribution-API.md
@@ -108,14 +108,14 @@ partial interface HTMLSlotElement {
 
 Step 4 is required because we have to re-calculate [assigned nodes] of every slot in the tree at this timing.
 
-Note: The detail is explained [later](#4.2.2.3.-Finding-slots-and-slotables), however, it would be worth noting that `manuallyAssignedNodes` is not used as
+Note: It would be worth noting that `manuallyAssignedNodes` is not used as
 [assigned nodes] as is. You can think of `slot.assign(sequence<Node> nodes)` as telling the engine a set of "candidate nodes" from which [assigned nodes] are constructed.
 
 [assign slotables for a tree]: https://dom.spec.whatwg.org/#assign-slotables-for-a-tree
 
 ## Changes to DOM Standard
 
-## [ShadowRootInit](https://dom.spec.whatwg.org/#dictdef-shadowrootinit)
+### [Shadow Root Init](https://dom.spec.whatwg.org/#dictdef-shadowrootinit)
 
 ``` webidl
 dictionary ShadowRootInit {
@@ -125,7 +125,7 @@ dictionary ShadowRootInit {
 };
 ```
 
-## [ShadowRoot](https://dom.spec.whatwg.org/#shadowroot) 
+### [Shadow Root](https://dom.spec.whatwg.org/#shadowroot) 
 
 ``` webidl
 [Exposed=Window]
@@ -139,8 +139,6 @@ interface ShadowRoot : DocumentFragment {
 enum ShadowRootMode { "open", "closed" };
 enum SlotAssignmentMode { "manual", "auto" };
 ```
-
-enum ShadowRootMode { "open", "closed" };
 
 ### [concept node insert](https://dom.spec.whatwg.org/#concept-node-insert)
 
--- a/proposals/Imperative-Shadow-DOM-Distribution-API.md
+++ b/proposals/Imperative-Shadow-DOM-Distribution-API.md
@@ -3,7 +3,7 @@
 Authors: yuzhehan@chromium.org, hayato@google.com
 
 This is a strawperson proposal for Imperative Shadow DOM distribution API.
-For more context, please see [the WhatWG issue](https://github.com/whatwg/html/issues/3534).
+For more context, please see [the WHATWG HTML Standard issue](https://github.com/whatwg/html/issues/3534).
 
 ## Motivation
 
--- a/proposals/Imperative-Shadow-DOM-Distribution-API.md
+++ b/proposals/Imperative-Shadow-DOM-Distribution-API.md
@@ -20,15 +20,15 @@ Some people would see this as a kind of *ugly* markup.
 Shadow DOM v1 can't explain how `<summary>/<details>` elements can be implemented on the top of the current Web Components technology stack, given that `<details>` element doesn't need slot= attribute.
 Blink has a special logic for some built-in elements to control node-to-slot mapping.
  
-#### Case 2: Can’t slot based on condition. 
-``` 
-<shadow-host num-to-show=”2”>
-   <div slot=”slot1”></div>
-   <div slot=”slot1”></div>
-   <div slot=”slot1”></div>
-</shadow-host>
+#### Case 2: Slot based on condition. 
+```html
+<custom-tab show-panel=”2”>
+   <tab-panel></tab-panel>
+   <tab-panel></tab-panel>
+   <tab-panel></tab-panel>
+</custom-tab>
  ```
-The second issue is that component creators can’t change the slotting behavior based on condition. A component that wants to display a subset of slottable based on a property can’t easily be done via declarative slotting.
+The second issue is that component creators can't change the slotting behavior based on condition. In the above markup, the `<custom-tab>` component can't implement which `<tab-panel>` to show based on its attribute `show-panel`.
 
 ## Imperative Slotting API
 
@@ -36,16 +36,41 @@ The imperative slotting API allows the developer to explicitly set the [assigned
 
 [assigned nodes]: https://dom.spec.whatwg.org/#slot-assigned-nodes
 
-An `HTMLSlotElement` gets a new API, called `assign()` (tentative name), and a new property, `manually-assigned-nodes`,
-which is an ordered list of nodes. Unless stated otherwise, it is empty.
+An `HTMLSlotElement` gets a new API, called `assign(sequence<Node> nodes)` (tentative name). Nodes are an ordered list of `slot's` host light DOM children. 
+ 
+#### Example Usage
+The following code demonstrate how to use the new API to solve [Case 2](#Case-2:-Slot-based-on-condition) of the Motivation section above.
 
-Basically, `slot.assign(sequence<Node> nodes)` sets the slot's `manually-assigned-nodes` to `nodes`. See the later section for details.
-
-- *manually-assigned-nodes* is an internal field. It is write-only. Users cannot read the value directly.
-
-- *manually-assigned-nodes* are different than [assigned nodes]. They are candidates for slot assignments. However, we check each node's slotable validity. If an invalid node is detected, we throw an exception in `slot.assign()` and clear `manually-assigned-nodes`. If the validity check is successful, the browser recalculates [assigned nodes] later, based on `manually-assigned-nodes`. The calculated [assigned nodes] are then observable.
-
-See examples below.
+```js
+class CustomTab extends HTMLElement {
+    constructor() {
+        super();
+        const template = `
+            <div class="custom-tab">
+                <slot></slot>
+            </div>`;
+        const shadowRoot = this.attachShadow({mode: 'open', slotAssignment: 'manual'});
+        const div = document.createElement('div');
+        div.innerHTML = template;
+        shadowRoot.appendChild(div);
+    }
+    connectedCallback() {
+        if (!this._observed) {
+           const target = this;
+           const showPanel = this.getAttribute('show-panel');
+           this._slot = this.shadowRoot.querySelector("slot");
+           const observer = new MutationObserver(function(mutations) {
+                const panels = target.querySelectorAll('tab-panel');
+                if (panels.length && showPanel > 0 && showPanel <= panels.length ) {
+                  target._slot.assign([panels[showTab-1]]);
+                } 
+            });
+            observer.observe(this, {childList: true});
+            this._observed = true;
+        }
+    }
+}
+```
 
 ## Declarative API and Imperative API should be exclusively used in each shadow tree
 
@@ -56,7 +81,7 @@ Thus, we don't allow mixing the declarative and imperative APIs in the same shad
 to use an imperative API for each shadow tree. They do this with a new "slotAssignment" parameter on attachShadow():
 
 ```js
-const sr = attachShadow({ mode: ..., (optional) slotAssignment: 'manual' });
+const sr = attachShadow({ mode: 'open', slotAssignment: 'manual' });
 ```
 
 Here, "manual" means "we support only imperative APIs for the shadow tree". If no calls to `slot.assign()` are made, then [assigned nodes] will be left empty.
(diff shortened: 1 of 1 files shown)


proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -1,48 +1,70 @@
...
+## Declarative API and Imperative API should be exclusively used in each shadow tree
 
-- *manually-assigned-nodes* is an internal field. It is write-only. Users cannot read the value directly.
+Mixing declarative API and imperative API would be troublesome and can cause confusion for web developers.
+We *can* invent complex rules. However, no one wants to remember complex rules. Also, supporting both in the same shadow tree would make a browser engine complex.
 
-- *manually-assigned-nodes* are different than [assigned nodes]. For example, users can pass a node which is not host's children. We don't throw an exception even if such a *invalid* node is passed, however, *invalid* node is never selected as [assigned nodes]; the engine recaluculate [assigned nodes] later, based on *manually-assigned-nodes*. Invalid nodes in *manually-assigned-nodes* are simply ignored. The caculataed [assigned nodes] are only observable.
+Thus, we don't allow mixing the declarative and imperative APIs in the same shadow tree. Web developers must explicitly opt-in
+to use an imperative API for each shadow tree. They do this with a new "slotting" parameter on attachShadow():
```
JanMiksovsky: Most of this proposal (and the existing spec) talks about _assignment_ rather than "slotting". Recommend reflecting that by changing the option to `assignment` (or maybe `slotAssignment`, etc.):

```this.attachShadow({ assignment: 'manual' })```

That makes the parameter's connection to the `assign()`, `assignedNodes`, and `assignedElements` more direct.

Comment (yuzhe-han) [reply proposals/Imperative-Shadow-DOM-Distribution-API.md]: Makes sense.  I like the slotAssignment parameter because it's more descriptive of what type of assignment it is.

Comment (yuzhe-han) [reply proposals/Imperative-Shadow-DOM-Distribution-API.md]: Updated doc to  use 'slotAssignment' instead.

File: proposals/Imperative-Shadow-DOM-Distribution-API.md [hunk 2]
```diff
@@ -1,48 +1,70 @@
 # Imperative Shadow DOM Distribution API
 
-hayato@google.com
+Authors: yuzhehan@chromium.org, hayato@google.com
 
-The straw-man proposal for Imperative Shadow DOM distribution API.
-The context is: [https://github.com/whatwg/html/issues/3534](https://github.com/whatwg/html/issues/3534)
+This is a strawman proposal for Imperative Shadow DOM distribution API.
```
domenic:
```suggestion
This is a strawperson proposal for Imperative Shadow DOM distribution API.
```

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -1,48 +1,70 @@
...
+For more context, please see [the WhatWG issue](https://github.com/whatwg/html/issues/3534).
```
domenic:
```suggestion
For more context, please see [the WHATWG HTML Standard issue](https://github.com/whatwg/html/issues/3534).
```

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -1,48 +1,70 @@
...
-Web developers have to show their *opt-in* to use imperative API for each shadow tree.
-
-A shadow root has an associated *slotting*. Web developers can set shadow root's *slotting* to *manual* by specifying it in attachShadow:
-
-```js
-const sr = attachShadow({ mode: ..., (optional) slotting: 'manual' })
+One of the drawbacks of Shadow DOM v1, when compared to Shadow DOM v0, is that web developers have to specify slot= attribute for every shadow host's children (except for elements for the *default* slot).
+ 
+#### Case 1:  Slot attributes is required in markup.
 ```
```
domenic:
```suggestion
```html
```

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -1,48 +1,70 @@
...
+  <div slot=slot1></div>
+  <div slot=slot2></div>
+</shadow-host>
+ ```
+Some people would see this as a kind of *ugly* markup.
+Shadow DOM v1 can't explain how `<summary>/<details>` elements can be implemented on the top of the current Web Components technology stack, given that `<details>` element doesn't need slot= attribute.
+Blink has a special logic for some built-in elements to control node-to-slot mapping.
+ 
+#### Case 2: Can’t slot based on condition. 
+```
```
domenic: Style issue: Use `html` after the three backticks, and fix all the curly quotes here to be normal quotes.

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -1,48 +1,70 @@
...
+ 
+#### Case 2: Can’t slot based on condition. 
+``` 
+<shadow-host num-to-show=”2”>
+   <div slot=”slot1”></div>
+   <div slot=”slot1”></div>
+   <div slot=”slot1”></div>
+</shadow-host>
+ ```
+The second issue is that component creators can’t change the slotting behavior based on condition. A component that wants to display a subset of slottable based on a property can’t easily be done via declarative slotting.
```
domenic: It'd be good to explain the connection to the above markup. Something like "So, there is no way to implement the `num-to-show` attribute in the above example.

It'd be even better if the example were more realistic, e.g. using an attribute like "tab" instead of "num-to-show" and a tag name like "tab-panel" instead of "shadow-host".

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -1,48 +1,70 @@
...
+
+## Imperative Slotting API
+
+The imperative slotting API allows the developer to explicitly set the [assigned nodes] for a slot element.
 
-The *manual* means "we support only imperative APIs for the shadow tree".
-The default is "we support only declarative API for the shadow tree".
+[assigned nodes]: https://dom.spec.whatwg.org/#slot-assigned-nodes
 
+An `HTMLSlotElement` gets a new API, called `assign()` (tentative name), and a new property, `manually-assigned-nodes`,
```
domenic: Properties in JavaScript are usually named in camel case, e.g. `slot.manuallyAssignedNodes`. Maybe you mean "associated value", instead of property?

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -1,48 +1,70 @@
...
+``` 
+<shadow-host num-to-show=”2”>
+   <div slot=”slot1”></div>
+   <div slot=”slot1”></div>
+   <div slot=”slot1”></div>
+</shadow-host>
+ ```
+The second issue is that component creators can’t change the slotting behavior based on condition. A component that wants to display a subset of slottable based on a property can’t easily be done via declarative slotting.
+
+## Imperative Slotting API
```
domenic, line 33: This section contains no code examples, and jumps straight into a discussion of the spec mechanisms and internal ways in which something will be implemented. See https://github.com/w3ctag/w3ctag.github.io/blob/master/explainers.md for a bit more on how to write a good explainer, in particular by including example code showing the API solving problems---perhaps the two problems you've introduced in the beginning.

I would:

- Move the discussion of internal concepts and implementation details to an appendix near the end.
- Add some basic examples (not as long or detailed as the ones you have below) showing how this API solves the problem.

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -1,48 +1,70 @@
...
+Mixing declarative API and imperative API would be troublesome and can cause confusion for web developers.
+We *can* invent complex rules. However, no one wants to remember complex rules. Also, supporting both in the same shadow tree would make a browser engine complex.
 
-- *manually-assigned-nodes* are different than [assigned nodes]. For example, users can pass a node which is not host's children. We don't throw an exception even if such a *invalid* node is passed, however, *invalid* node is never selected as [assigned nodes]; the engine recaluculate [assigned nodes] later, based on *manually-assigned-nodes*. Invalid nodes in *manually-assigned-nodes* are simply ignored. The caculataed [assigned nodes] are only observable.
+Thus, we don't allow mixing the declarative and imperative APIs in the same shadow tree. Web developers must explicitly opt-in
+to use an imperative API for each shadow tree. They do this with a new "slotAssignment" parameter on attachShadow():
 
-See also the Example 3 later.
+```js
+const sr = attachShadow({ mode: ..., (optional) slotAssignment: 'manual' });
```
domenic: This isn't valid JavaScript syntax, so it's kind of confusing. I suggest:

```suggestion
const sr = attachShadow({ mode: 'open', slotAssignment: 'manual' });
```

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -54,36 +76,55 @@ partial interface HTMLSlotElement {
...
+2. Validate each node in nodes that its node.parentNode == slot.rootNode.host. Throw if not equal.
+3. Set the slot's `manually-assigned-nodes` to `nodes`.
+4. Run [assign slotables for a tree] with slot’s tree.
 
-step 2 is required because we have to re-calculate [assigned nodes] of every slots in the tree at this timing.
+Step 4 is required because we have to re-calculate [assigned nodes] of every slot in the tree at this timing.
 
-Note: The detail is explained later, however, it would be worth noting that *manually-assigned-nodes* is not used as
-[assigned nodes] as is. You can think that `slot.assign(sequence<Node> nodes)` tell the engine *candidate nodes* from where [assigned nodes] are constructed.
+Note: The detail is explained later, however, it would be worth noting that `manually-assigned-nodes` is not used as
```
domenic: Where is "later"? Can you link to it?

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -54,36 +76,55 @@ partial interface HTMLSlotElement {
...
 
-# Changes to DOM Standard
+## Changes to DOM Standard
 
-## ShadowRootInit
+### ShadowRootInit
 
 
 ``` webidl
 ShadowRootInit {
```
domenic: It would be good to rewrite this closer to valid Web IDL.

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -93,35 +134,33 @@ Other steps don't need to be updated from the standard's perspective, I think, t
...
-├──/shadowroot (slotting=manual)
+├──/shadowroot (slotAssignment=manual)
 │   ├── slot1
 │   └── slot2
 ├── A
 └── B
-
 ```
 
 ``` javascript
```
domenic: The "== means ArrayEquals" is very confusing here since it means is isn't JavaScript code. It would be good to replace this all with `assert_array_equals`.

domenic (commented):
Exciting to see this being worked on! Some suggestions on how to make it better.

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -54,36 +76,55 @@ partial interface HTMLSlotElement {
...
-[To find a slot] for a given slotable *slotable* and an optional open flag (unset unless stated otherwise), run these steps:
+    1. Adopt node into parent’s node document.
+
+    2. If child is null, then append node to parent’s children.
+
+    3. Otherwise, insert node into parent’s children before child’s index.
+    
+    4. [Update Step] If parent is a shadow host, its shadow's `slotAssignment` is not "manual" and node is a slotable, then assign a slot for node.
+
+    5. [New Step] If parent's root is a shadow root, shadow's `slotAssignment` is "manual", and parent is a slot, throw exception.
```
bicknellr: This prevents slots in shadow roots with manual distribution from having fallback content. Is that intended?

proposals/Imperative-Shadow-DOM-Distribution-API.md
yuzhe-han: Updated to use html and fixed all curly quotes to use normal quotes instead.

proposals/Imperative-Shadow-DOM-Distribution-API.md
yuzhe-han: Thanks for the suggested example. Updated my example to be a better case. Also added an example implementation to showcase the new API.

proposals/Imperative-Shadow-DOM-Distribution-API.md
yuzhe-han: Good call out. I've changed this variable to be JS compliant. It's a implementation detail. I've added a appendix section at the bottom to explain the usage of this variable.

proposals/Imperative-Shadow-DOM-Distribution-API.md
yuzhe-han: Got it. Example addedd and moved the internal concepts to appendix at the bottom.

proposals/Imperative-Shadow-DOM-Distribution-API.md
yuzhe-han: Update

proposals/Imperative-Shadow-DOM-Distribution-API.md
yuzhe-han: Updated.

proposals/Imperative-Shadow-DOM-Distribution-API.md
yuzhe-han: Updated with WebIDL

proposals/Imperative-Shadow-DOM-Distribution-API.md
yuzhe-han: Removed.

proposals/Imperative-Shadow-DOM-Distribution-API.md
yuzhe-han: bicknellr@,

Yes, it's intended, but not sure if the spec wording is correct. Or if it needs to throw an exception vs no-op.

To me, fallback content makes it confusing when working in slot assignment manual mode. Slot assignment should be absolute and final in manual mode.  See comment from TPAC: https://github.com/whatwg/html/issues/3534#issuecomment-537802687

In that case, when developers call slot.assign([]), the slot should be cleared and have no content. But if we allow failback content, then content can still appear inside the slot even thou developer specifically want to clear it out. 

This is my reasoning for update the spec. Let me know if you think I'm overlooking something.

thanks

yuzhe-han: bicknellr@,

After some considering and discussing it with masonfreed@, we decided it's better to preserve the fallback content behavior. So for a shadow root, regardless of whether slotAssignment is manual or auto, the fallback behavior is identical. 

I've since removed this section in the explainer doc. Thanks again for your review.

bicknellr: Thanks for taking another look at this. I think that's the best option given that it makes the behavior the same in both cases and that slots' fallback content is very likely controlled by the same entity that would assign nodes.

rniwa:
Looks like we never got to discuss this topic at this week's F2F.

@yuzhe-han: would you be interested in presenting / championing this proposal in the next virtual F2F if we ever have such a thing?

yuzhe-han:
> Looks like we never got to discuss this topic at this week's F2F.
> 
> @yuzhe-han: would you be interested in presenting / championing this proposal in the next virtual F2F if we ever have such a thing?

@rniwa: Sure. Would love a chance to present and meet you all.

yuzhe-han:
@domenic,

Hopefully, I've addressed all your issues. Can you please take another look?

Thanks.

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -54,210 +101,261 @@ partial interface HTMLSlotElement {
...
 
 ``` webidl
-ShadowRootInit {
-   ...
-   (optional) sloting: 'manual'|'auto' //  (if omitted, it is 'auto');
-}
+dictionary ShadowRootInit {
+  required ShadowRootMode mode;
+  boolean delegatesFocus = false;
+  SlotAssignmentMode slotAssignment;   // optional. (defaults to 'auto')
```
domenic:
```suggestion
  SlotAssignmentMode slotAssignment = "auto";
```

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -1,48 +1,95 @@
...
+  <div slot=slot2></div>
+</shadow-host>
+ ```
+Some people would see this as a kind of *ugly* markup.
+Shadow DOM v1 can't explain how `<summary>/<details>` elements can be implemented on the top of the current Web Components technology stack, given that `<details>` element doesn't need slot= attribute.
+Blink has a special logic for some built-in elements to control node-to-slot mapping.
+ 
+#### Case 2: Slot based on condition. 
+```html
+<custom-tab show-panel=”2”>
```
domenic:
```suggestion
<custom-tab show-panel="2">
```

proposals/Imperative-Shadow-DOM-Distribution-API.md
```diff
@@ -1,48 +1,95 @@
...
+        const template = `
+            <div class="custom-tab">
+                <slot></slot>
+            </div>`;
+        const shadowRoot = this.attachShadow({mode: 'open', slotAssignment: 'manual'});
+        const div = document.createElement('div');
+        div.innerHTML = template;
+        shadowRoot.appendChild(div);
+    }
+    connectedCallback() {
```
domenic, line 60: You might want to add a comment explaining that a production example would also use attributeChangeCallback to watch for changes to the `show-panel` attribute and re-run this code.

You could also just update the example to do that, but I'm not sure how much complexity it would add; perhaps it would be distracting.

domenic (approved):
LGTM with a couple nits; thank you!

I'd like to give you write access to this repo, so that you can merge on your own after fixing those nits, but I'm not sure who has permissions to do that. Maybe @plehegar can help.

proposals/Imperative-Shadow-DOM-Distribution-API.md
yuzhe-han: I've updated my example to include attributeChangedCallback.
Thanks.

PR #878, opened 2020-05-15
Update HTML Modules explainer phrasing around allowed element types

trusktr:
I modified 

> `<script>` elements of `type="module"` (non-module scripts will cause HTML Module creation to fail)

to 

> `<script>` elements of `type="module"` (non-module scripts will cause HTML Module creation to fail) and `<template>` elements

Is this correct? It seems to be what the explainer hints at.


trusktr:
Nevermind, I was confused. At first I thought the original phrase meant only a `<script>` tag and no other tags could be in the HTML Module (which would be fairly useless). But I now saw other examples with arbitrary DOM elements besides the `<script>`.

New pull request: https://github.com/w3c/webcomponents/pull/880

PR #888, opened 2020-07-11
start to improve the index page, add proposals section

trusktr:


annevk:
I think we should remove this file instead. I think README.md will then show up, which is at least somewhat maintained.

ashly1111111 (approved):

Issue #884, opened 2020-07-10
[declarative-custom-elements] how does the class get associated with the definition


trusktr:
In the following,

```html
<definition name="my-element">
    <template>....</template>
    <script type="module">
        class MyElement extends HTMLElement {
            #template;
            constructor(state, ...args) {
                super(state, ...args);
                #template = customElements.attachTemplateAsShadow(this, state);
            }
            render(state) {
                #template.update(state);
            }
        }
    </script>
</definition>
```

how does the engine know to define the `<my-element>` element using the `MyElement` class?

How does `type="module"` come into play here? What happens if type=module were not included?

What happens if I write two classes in the script tag? F.e.

```html
<definition name="my-element">
    <template>....</template>
    <script type="module">
        class MyElement extends HTMLElement {
            // ...
        }
        class MyOtherElement extends HTMLElement {
            // ...
        }
    </script>
</definition>
```

Or is the example in the proposal missing an `export`? Should it be

```html
<definition name="my-element">
    <template>....</template>
    <script type="module">
        export default class MyElement extends HTMLElement {
            // ...
        }
    </script>
</definition>
```

?

Happy to propose changes to the proposal file once this is cleared up.

trusktr:
I'll assume that the proposal was missing the `export` part.

What happens if there are multiple script tags?

```html
<definition name="my-element">
    <template>....</template>
    <script type="module">
       export default class MyElement extends HTMLElement {
           // ...
       }
    </script>
    <script type="module">
       export default class MyOtherElement extends HTMLElement {
           // ...
       }
    </script>
</definition>
```

Should the `<definition>` accept and execute only the first script tag? Or maybe execute any script tags like normal, but take the export only from the first script tag?

Maybe the following is for another spec, but it seems there should be a way, in general, to grab the exports of a type=module script tag, f.e.

```html
<script type="module">
   export default class MyOtherElement extends HTMLElement {
       // ...
   }
</script>
<script>
  const s = document.querySelector('[type=module]')

  // Similar to ES import():
  s.import().then(m => console.log(m.default)) // logs MyOtherElement class
</script>
```

Or maybe also something like

```html
<script type="module" name="MyOtherElement">
   export default class MyOtherElement extends HTMLElement {
       // ...
   }
</script>
<script>
  const s = document.querySelector('[type=module]')

  // Using ES import():
  import('script:MyOtherElement').then(m => console.log(m.default)) // logs MyOtherElement class
</script>
```

[...]

PR #897, opened 2020-10-20
Declarative components proposal

emanuelhfarias:
Today to create a web component we have to write a piece of JavaScript code just to register a custom element. Web component needs to give a next step. If we want people to use it more we should let them create small components without friction. A small component is just a custom element for encapsulating template and/or style, no JS required.

A declarative way to define custom elements appears to solve the problem above. Some proposals has already been made, but apparently, none has evolved.

I've been doing some experiments on this topic and I've created a polyfill for the tag `<component>` which can be the wrapper for declarative define a custom element.

Imagine you can define a custom element like this:

```html
<component name="alert-box" extends="p">
  <style>
    p {
      color: #FFF;
      background-color: #3778e5;
      border-radius: 6px;
      padding: 5px 15px;
    }
  </style>
</component>

<alert-box>some cool text!</alert-box>
```

You can find several live examples using the polyfill [here](https://tag-components.vercel.app).

I would like to invite some people to join the conversation:
@ebidel, @rniwa, @trusktr

What do you think?

Thanks

### DIFF
--- a/proposals/Declarative-Components.md
+++ b/proposals/Declarative-Components.md
@@ -1,0 +1,102 @@
+# Declarative Components
+
+### Problem
+Today to create a web component we have to write a piece of JavaScript code just
+to register a custom element. Web component needs to give a next step. If we
+want people to use it more we should let them create small components without
+friction.
+
+A declarative way to define custom elements appears to solve the problem above.
+Some proposals has already been made, but apparently, none has evolved.
+
+### Proposal
+
+This proposal introduces the new tag `<component>` to HTML.
+
+A component may be composed of three parts: template, style and script.
+None of them are required.
+
+```html
+<component name="badge-number" extends="span">
+  <template>
+  </template>
+
+  <style>
+  </style>
+
+  <script>
+  </script>
+</component>
+```
+
+Today, browsers automatically creates a `#document-fragment` for every `<template>` tag it finds.
+
+This proposal suggest browsers to:
+
+1. create a `#document-fragment` for every `<component>` element it finds.
+2. automatically register a custom element as soon as it parses a `<component>` element.
+
+Registering a component have four steps:
+1. register the template
+2. register the root element
+3. register the style
+4. register the script
+
+#### The template
+A web component can have only one `<template>` tag inside. It is not required.
+
+#### The Root Element
+If the component `extends` from another element and does not have a `<template>` it should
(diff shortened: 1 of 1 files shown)


justinfagnani:
Declarative custom elements is a topic that's come up a lot over the years, and we have a strawman proposal here: https://github.com/w3c/webcomponents/blob/gh-pages/proposals/Declarative-Custom-Elements-Strawman.md I think the current implicit status is that we need to get through Template Instantiation, and then a built-in expression syntax that we could use in HTML, before doing a lot more on declarative custom elements, though maybe we could work on them in parallel and maybe the two proposals would lend each other momentum.

I'm very glad to see more interest and new proposals here, though at this point I personally think we'll need more discussions in issues and meetings before another PR, and don't think we should have two proposals in the proposals folder. A new proposal PR right now is likely to have deeps discussion on specifics in review comments that should be had in separate issues.

But a few things on this proposal that I think are worth mentioning:

1. To be viable, I think declarative custom elements need dynamic templates, we'll need lots of consideration over the data model they have access to and the security implications.
2. `<style>` can't just be a child element outside of a template, at least without a new value for `type`, as it'll apply to the document. We need a solution for adopted style sheets with declarative shadow DOM, so hopefully we can share a solution there with something like `<style type="adoptable">`.
3. I don't think we should special case a "root" element in the shadow root, since we haven't yet. `:host()` isn't inconvenient, IMO. `:host([foo=bar])` is barely different than `div[foo=bar]`. CSS custom state and Accessibility Object Model should alleviate other pressures to have a root element.
4. For upgrades with a `<script>` tag we shouldn't need to require anything dynamic wrt the base class. Elements can extend `HTMLElement` and upgrade as they do now. For association between a script and a declarative definition we should be able to either use a default export for inline scripts, or have a new API to upgrade a declaratively defined element. We have an issue here now: https://github.com/w3c/webcomponents/issues/884

I think it'd be great to open more issues and discuss this more there and in upcoming meetings.

emanuelhfarias:
@justinfagnani thanks for clarify the state we are on declarative custom elements.

I'm closing this PR and I'll read all other discussions and proposals. Probably this way I can contribute more.

One thing though, I miss live examples. All discussions I see so far just show one or two examples.

It would be nice if we have a place to play with declarative components from the discussions.

The [website](https://tag-components.vercel.app/) I did for this proposal have this intention. Maybe showing off what can be done with declarative components will get more people's attention.

PR #915, opened 2021-03-12
[scoped-registries] Describe interaction with declarative shadow DOM

### LINKED ISSUE #914: Interaction with declarative shadow DOM
We need to specify how scoped custom element registries interact with declarative shadow DOM.

The fundamental behavior we need is for a declarative shadow root to _not_ use the global registry when it's going to be used with a scoped registry once its host is defined. Then we need the host to be able to assign a scoped registry to a shadow root that's already been created.

This is accomplishable with an [attribute](https://github.com/mfreed7/declarative-shadow-dom/blob/master/README.md#additional-arguments-for-attachshadow) that maps to an `attachShadow()` option to opt-out of the global registry, without yet having a scoped registry to use. In `attachShadow()` options, we can allow `registry` to be undefined. In declarative shadow DOM, we can introduce an attribute, like `shadowrootregistry`:

```
<template shadowroot="open" shadowrootregistry="">
  <some-scoped-element></some-scoped-element>
</template>
```

Next, we need to be able to add the ability for a scoped registry to be set on an existing ShadowRoot. Possibly like:

```ts

const registry = new CustomElementRegistry();

class extends HTMLElement {
    #internals = null;
    constructor() {
      super();
      this.#internals = this.attachInternals();
      if (this.#internals.shadowRoot) {
        this.#internals.shadowRoot.registry = registry;
      } else {
        this.attachShadow({mode: 'open', registry});
      }
    }
 }
```

This means we need to clarify whether the registry can change over time, or whether it can only be set on declarative shadow roots that have been opted out of the global registry. A registry changing over time is something that has also come up in https://github.com/WICG/webcomponents/issues/907 (moving shadow roots between documents).

justinfagnani:
Fixes #914 

This adds a `shadowrootregistry` attribute for `<template shadowrootmode>` and `ShadowRoot.prototype.scopedRegistry`, as [discussed in the String 2023 F2F meeting](https://docs.google.com/document/d/1XWp4ndl5bthoxhnqR9Ly1SYvYcmXw1eTrappoMXsuH4/edit#bookmark=id.y2d7g72l1qm).

### DIFF
--- a/proposals/Scoped-Custom-Element-Registries.md
+++ b/proposals/Scoped-Custom-Element-Registries.md
@@ -121,19 +121,25 @@ This poses a limitation for authors trying to use the constructor to create new 
 
 ## Interaction with Declarative Shadow DOM
 
-[Declarative shadow DOM](https://github.com/mfreed7/declarative-shadow-dom/blob/master/README.md) allows HTML to construct shadow roots for elements from `<template>` elements with a `shadowroot` attribute.
+[Declarative shadow DOM](https://github.com/mfreed7/declarative-shadow-dom/blob/master/README.md) allows HTML to construct shadow roots for elements from `<template>` elements with a `shadowrootmode` attribute.
 
 Since these shadow roots are not created by a host calling `attachShadow()`, the host doesn't have a chance to pass in a scoped custom element registry. If a host is using a scoped registry, we need to force the declarative shadow root to not use the global registry and instead leave custom elements un-upgraded until the host can create and assign the correct registry.
 
 To do this we add a `shadowrootregistry` attribute, according to the [declarative shadow root explainer section on additional `attachShadow()` options](https://github.com/mfreed7/declarative-shadow-dom/blob/master/README.md#additional-arguments-for-attachshadow). This attribute causes the declarative shadow root to have no registry associated with it at all:
 
 ```html
-<template shadowroot="open" shadowrootregistry="">
-  <some-scoped-element></some-scoped-element>
-</template>
+<my-element>
+  <template shadowrootmode="open" shadowrootregistry="">
+    <some-scoped-element></some-scoped-element>
+  </template>
+</my-element>
 ```
 
-When the host is defined and upgrades with an existing shadow root, it can assign the registry:
+The shadow root created by this HTML will have a `null` registry, and be in a "awaiting scoped registry" state that allows the registry to be set once after creation.
+
+To identify this "no registry, but awaiting one" state, ShadowRoot will have a `scopedRegistry` boolean property. `scopedRegistry` will set to `true` for all ShadowRoots with a scoped registry, or awaiting a scoped registry. Code can tell that ShadowRoot has an assignable `registry` property if `root.registry === null && root.scopedRegistry === true`.
+
+Host elements with declarative scoped registries can assign the correct registry during upgrades, like so:
 
 ```ts
 const registry = new CustomElementRegistry();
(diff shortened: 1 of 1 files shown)


justinfagnani:
@rniwa @mfreed7 I updated this PR with [what we resolved from last weeks meeting](https://docs.google.com/document/d/1XWp4ndl5bthoxhnqR9Ly1SYvYcmXw1eTrappoMXsuH4/edit#bookmark=id.y2d7g72l1qm), by adding `ShadowRoot.prototype.scopedRegistry`.

Can one of you review? Thanks!

proposals/Scoped-Custom-Element-Registries.md
```diff
@@ -119,6 +119,51 @@ As a result, it must limit constructors by default to only looking up registrati
...
+<my-element>
+  <template shadowrootmode="open" shadowrootregistry="">
+    <some-scoped-element></some-scoped-element>
+  </template>
+</my-element>
+```
+
+The shadow root created by this HTML will have a `null` registry, and be in a "awaiting scoped registry" state that allows the registry to be set once after creation.
+
+To identify this "no registry, but awaiting one" state, ShadowRoot will have a `scopedRegistry` boolean property. `scopedRegistry` will set to `true` for all ShadowRoots with a scoped registry, or awaiting a scoped registry. Code can tell that ShadowRoot has an assignable `registry` property if `root.registry === null && root.scopedRegistry === true`.
```
rniwa: We should probably have "shadowRoot.customElements" instead of "shadowRoot.scopedRegistry" to be consistent with the global registry.

justinfagnani: I think that would be `shadowRoot.customElements` instead of `shadwoRoot.registry`?

rniwa: right.

mfreed7: Hmm, I find `shadowRoot.customElements` a bit harder to understand, as it's defined. In other words, this boolean is true "when the shadowRoot has a scoped registry, or is waiting for a scoped registry". I feel like "registry" should be at least somewhere in the name of the property, right?

mfreed7 (approved):

proposals/Scoped-Custom-Element-Registries.md
justinfagnani: @rniwa meant that `registry` should be named `customElements` to match the global `window.customElements`.

`shadowRoot.scopedRegistry` would still be the boolean, if that's agreeable. (I don't think we do `is` or `has` prefixes in the DOM, correct?)

mfreed7: Oh! Thanks, I missed that. I agree with renaming `shadowRoot.registry` to `shadowRoot.customElements`, which is exactly what you both wrote, but I failed to read correctly.

The DOM does use `is` sometimes, e.g. `isConnected`, `isTrusted`, `isMap`, `isContentEditable`, etc. The `has` prefix (which seems like what we'd want here) is also used, though less so. For example, `hasBeenActive`. If `has` is acceptable, I'd definitely lobby for renaming `shadowRoot.scopedRegistry` to `shadowRoot.hasScopedRegistry`.

justinfagnani: I'll change to `hasScopedRegistry` in this PR and we can see what other reviewers think.

As for `registry` vs `customElements`: that effects the argument to `attachShadow()` too. I think I should make a separate PR for both so the change is more obvious?

EisenbergEffect:
What is the status of this WRT the latest scope registry implementations in Chromium and WebKit? Is `shadowrootregistry=""` on `<template>` supported as part of that work? And do we have a way to associate the scoped registry later on, such as through a `registry` or `customElements` property somewhere?

I'm playing around with things now, and since server rendering with DSD is an important strategy for me, I'm trying to understand how best to integrate that with scoped registries, given the recent progress in browser engines.

rniwa:
> What is the status of this WRT the latest scope registry implementations in Chromium and WebKit? Is `shadowrootregistry=""` on `<template>` supported as part of that work? And do we have a way to associate the scoped registry later on, such as through a `registry` or `customElements` property somewhere?
> 
> I'm playing around with things now, and since server rendering with DSD is an important strategy for me, I'm trying to understand how best to integrate that with scoped registries, given the recent progress in browser engines.

There is no support yet. We need to solidify the proposal to figure out what exactly we're gonna do here.

rniwa:
This issue is obsolete now that we've merged scooped custom element registry definitions into DOM & HTML specifications.

PR #932, opened 2021-06-11
Fix css-modules-v1-explainer.md

tomayac:
- Add missing `assert`s.
- Consistently use double quotes. 
- Add leading dot to import.


tomayac:
Closing since @dandclark deals with this in https://github.com/WICG/webcomponents/pull/931.

PR #948, opened 2022-01-03
Template instantiation: Fix typo

dy:
It's meant type "part" here


rniwa:
Hm... ipr bot is failing. I think you need to mark it as not IP substantial or something.

dy:
@rniwa someone authorized needs to mark it as non-substantive from ipr panel

PR #1023, opened 2023-07-31
[DOM Parts] Add new declarative syntax and update iterative proposal

tbondwilkinson:
This PR makes various changes against the current DOM parts proposal:

1. Adds a prospective declarative proposal. The basic syntax is `{{}}` for a self-contained part, `{{#}}{{/}}` for part with content. Parsing is enabled with a `parseparts` attribute on any DOM node.
2. `DocumentPartRoot` or `ChildNodePart` are now containers for other nested parts. Browsers will need to maintain the nested list of parts.
3. Cloning operation exists on `DocumentPartRoot` and `ChildNodePart` and returns a new `Part`.


proposals/DOM-Parts-Declarative.md
```diff
@@ -0,0 +1,128 @@
...
+    Email: <a id="link" href="{{}}">{{}}</a>
+  </section>
+</template>
+```
+
+This will create a `ChildNodePart` attached to `<h1>` with no content, an
+`AttributePart` connected to `href`, and a `ChildNodePart` connected to `<a>`
+with no content.
+
+A framework could fetch these parts use either a
```
rictic: Looks like the wording here was caught up in an edit

proposals/DOM-Parts-Declarative.md
```diff
@@ -0,0 +1,128 @@
...
+```html
+<template>
+  <section>
+    <h1 id="name">{{}}</h1>
+    Email: <a id="link" href="mailto:{{}}">{{}}</a>
+  </section>
+</template>
+```
+
+The `AttributePart` for `href` would have `value` equal to `['mailto:', '']`.
```
rictic: But the `value` is the mutable field that the user writes. Seems like we want to have a readonly array for the static strings, and then use `AttributePart#value` for the dynamic values

rictic: But the `value` is the mutable field that the user writes to to update value, isn't it? Seems like we want to have a readonly array for the static strings, and then use `AttributePart#value` for the dynamic values

proposals/DOM-Parts-Declarative.md
```diff
@@ -0,0 +1,128 @@
...
+Templating systems may need to serialize data about the nodes they are marking
+into the processing instructions. Or at the very least parts could be named so
+that they are easier to fetch.
+
+```html
+<div>{{email data="foo"}}</div>
+```
+
+This could be exposed on the imperative API to be consumed in JavaScript by
+application logic.
```
rictic, line 117: Feel free to punt, but have we considered giving the metadata as a string rather than a string array? People may want to use a variety of metadata encodings, e.g. JSON, and it'd be easier to just pass along the raw contents and let them split by spaces if they'd like to do that.

Related, but we'll want to specify how to escape `}}` inside of part syntax, and how to escape `{{` inside `parsepart` html

proposals/DOM-Parts-Declarative.md
```diff
@@ -0,0 +1,128 @@
...
+
+The `{{}}` and `{{#}}{{/}}` are reasonable DOM part markers, but it could be
+something else, or it could even be something that is declared as the
+placeholder.
+
+## Compatability
+
+It may be challenging to implement `<template>` parsing of `{{}}` for
+compatability reasons, so there may need to be restrictions such as only
+allowing this template syntax in a new API like `createTemplateWithParts()`.
```
rictic: Why not use a `parseparts` attribute on the `<template>` for this?

proposals/DOM-Parts-Imperative.md
```diff
@@ -8,43 +8,67 @@ A DOM part is represented by the `Part` interface and its sub interfaces.
...
+  constructor(
+      PartRoot root,
+      Element element,
+      DOMString qualifiedName,
+      optional DOMString? namespace = null,
+      optional PartInit init = {});
+
+  readonly attribute DOMString prefix;
+  readonly attribute DOMString localName;
+  readonly attribute DOMString namespaceURI;
```
rictic, line 39:
```suggestion
  readonly attribute DOMString namespaceURI;
  readonly attribute DOMString rawName;
```

Consider adding a field representing the attribute name as it was written in the original source, before it was lower cased.

proposals/DOM-Parts-Imperative.md
```diff
@@ -8,43 +8,67 @@ A DOM part is represented by the `Part` interface and its sub interfaces.
...
-    readonly attribute DOMString localName;
-    readonly attribute DOMString namespaceURI;
+  constructor(
+      PartRoot root,
+      Element element,
+      DOMString qualifiedName,
+      optional DOMString? namespace = null,
+      optional PartInit init = {});
+
+  readonly attribute DOMString prefix;
```
rictic, line 37:
```suggestion
  readonly attribute DOMString prefix;
  readonly attribute FrozenArray<DOMString> strings;
```

Consider adding a field representing the static strings that the values are interpolated in between, with `['', '']` as the default value.

proposals/DOM-Parts-Imperative.md
```diff
@@ -256,156 +212,67 @@ had initially contained `mailto:` before `{email}` but we could not capture this
...
-
-### Option 3. Introduce `AttributePartFragment`
-
-Unlike option 2, this creates `AttributePartFragment`s from `AttributePart`,
-meaning that `AttributePart` in option 3 plays the role of `AttributePartGroup`
-in option 2:
+Instead of `AttributePart` having a single string `value`, it could optionally
+take an `Array` that contains values that should be concatenated together. This
+allows updating individually parts of the attribute without needing to serialize
+the entire string.
```
rictic, line 222:
```suggestion
Instead of `AttributePart` having a single string `value`, it could instead
take an `Array` with values that will be interleaved with those from the `strings` array. 
```

If we go with the `strings` design, we might want to require that `values` be an array, to be interleaved with the `strings`. I suppose we could take a non-array value if there's only two strings, but in practice I expect template systems to just always pass through an array here.

proposals/DOM-Parts-Imperative.md
```diff
@@ -256,156 +212,67 @@ had initially contained `mailto:` before `{email}` but we could not capture this
...
-    return `${this.firstName} ${this.lastName}`;
-  }
-}
-
-const part = AttributePart(element, "title");
-part.value = new TitleAttributeValue();
-part.value.firstName = "Ryosuke";
-part.value.lastName = "Niwa";
+const part = AttributePart(element, "href", {
+  template: attribute`mailto: ${0}`,
```
rictic: I'm not sure if taking a template strings array here gives us any security benefits. The static and dynamic values will just be concatted together into a single string before they're set as an attribute.

Hm, on the other hand, if native elements got information about the static and dynamic values, that might give future content security policies the ability to set constraints that some parts of attribute strings must be constant. The `safevalues` library does something like that for high risk URL attributes like an iframe's `src`, where the origin must be statically determined, but the path can be dynamic.

Ok, in that case I think I see the value here.

proposals/DOM-Parts-Declarative.md
EisenbergEffect: That was my question as well.

mfreed7: +1 that should be enough for an opt-in.

EisenbergEffect:
I'd like to get a bit of clarification on syntax around parts with content **and** metadata. So, imagine a part that represents email content. It wants to provide both the binding expression metadata **and** the SSR'd value of the email. Would that look like this?...

```html
{{# email.expression.here}}john@doe.com{{/}}
```

It might be good to clarify this in the document.

proposals/DOM-Parts-Declarative.md
tbondwilkinson: Oh I was imagining we would replace the value field entirely. Because otherwise you could never update what otherwise looked like a static string?

tbondwilkinson: Duplicate comment

proposals/DOM-Parts-Declarative.md
tbondwilkinson: Oh yeah it's not clear, but the reason metadata is a string is not because we split the string, it's because you might actually have metadata in two places for the same part.

```
{{# metadata1}}foo{{/ metadata2}}
```

I'll call this out.

proposals/DOM-Parts-Declarative.md
tbondwilkinson: Updated the above section.

proposals/DOM-Parts-Imperative.md
tbondwilkinson: I think we need to discuss precisely what we want the `value` to look like.

proposals/DOM-Parts-Imperative.md
tbondwilkinson: I think we need to think more about the DX of this.

proposals/DOM-Parts-Declarative.md
rictic: Two motivating examples:

```html
  <div class="foo {{color}} bar"></div>
  <iframe src="https://partner.example.com/page?{{params}}">
```

In these cases, I think a user would expect that after updating these AttributePart's values, foo and bar would still be classes on the div, and the iframe would only load from the partner page.

If a library on top of template parts wants to update the attribute to a value that doesn't include the static strings (e.g. via a user level directive system, or when set to a magic value defined by the library), it could still do so by getting the element and attribute name off the part and calling imperative APIs.

sashafirsov:
**If you agree, please downvote this proposal.**

By imposing extra curly brace in attributes you are:
* adding extra 2 characters with no advantage behind
* increasing the reading complexity
* introducing additional syntax/parser into platform where enough parsers already exist
*  adding more diversity to syntax where existing ones can be reused

From prospective of normal developer and manager `{{ }}` is additional cost for training and skills.
I do not see any positive features related to such syntax. But there are bunch of draw backs. 


What is available on the platform as of now?
* JS template syntax. `${jsExpr}`. At least all who do the web development familiar with this syntax and parser is part of JS engine. The question of JS scope vs template context recognition is not related to template syntax, and in current state of things is allover industry where JS context been part of templates. Say in JSX. Not template-oriented though. Useful mostly in embedded context.
* XSLT/XPath syntax. `{xPathExpr}` is a shorter and still quite powerful. My favorite due to template-oriented features. 

Both syntax are available for DX as
* evaluation `eval()` and API accordingly
* well tested, time proven
* have a tons of validators as in CLI as in IDE level
* syntax colored

And all of this usefulness you want to dump in favor of custom solution?

rictic:
@sashafirsov We're not committed to the `{{}}` syntax, at the face to face it was agreed that it's a strawman, a placeholder for the time being. The goal is to get implementor and library feedback as to whether the semantics are sufficient and whether we can integrate it into HTML in a backwards compatible way that's not too disruptive to browsers' parsers.

sashafirsov:
@rictic , even if the syntax just a placeholder, the principle of attributes vs dom parts markup by same convention is important to separate on current discussion level. It can be done as in backward compatible as in breaking fashion. The parts within attribute already  have a built-into-browser solution: XSLT convention. The dom parts as well: either as custom element or `<xsl:value-of>`

I see the proposal as attempt to swing away from ability to reuse the existing platform features along with complimentary advantages without matching compensation.

The `<part>` is the good approach which can be generated by whatever syntax the build toolchain support whether it is JSX or mustache syntax. But the reverse ( as in proposal) is not that easy. 

There is a compromise which can/should come with separate proposal: the source DOM alias: 
`✂️=<part>` using either html entities extension of another mechanism. This way the mustache `{{` would be aliased to `<part>` or another string in parsing time. But if we would state the synthetic syntax upfront, that opportunity would not be an option.

rniwa:
Please take the discussion of your proposal elsewhere. This issue is specifically about template / DOM part API.

proposals/DOM-Parts-Declarative.md
tbondwilkinson: Okay, makes sense. I added a new `statics` attribute to `AttributePart` with declarative or imperative statics.

tbondwilkinson:
> I'd like to get a bit of clarification on syntax around parts with content **and** metadata. So, imagine a part that represents email content. It wants to provide both the binding expression metadata **and** the SSR'd value of the email. Would that look like this?...
> 
> ```
> {{# email.expression.here}}john@doe.com{{/}}
> ```
> 
> It might be good to clarify this in the document.

I added explicit mention of this in the declarative metadata section. How does that look?

tbondwilkinson:
To respond to Sasha briefly.

I think we are operating with the assumption that the majority of developers are working with frameworks, and I don't know of any popular frameworks that use XSLT syntax for templating. This has been true for many years. The choice of `{{}}` is not the standard for all templating languages, but it's close enough that we can use it as a placeholder for now before we survey more developers. XSLT does not seem to have developer interest, and adding new features that relied on it would seem to go against where the majority of developers are headed.

I think a number of us are also confident that we want a declarative syntax for template output, rather than just an imperative solution, which to me rules out template literal syntax in HTML. Template literals also have no concept of "start" "stop" ranges, and are merely insertion points for expressions, and we'd want our declarative syntax to be quite a bit more expressive.

I think `{}` syntax is still a candidate, but I don't think we are as interested in `<xsl:attribute>` or `<part>` syntax to mark specific attributes or nodes as parts. I feel strongly emojis are completely off the table for DOM part markers even as a "shorthand."

justinfagnani:
Another important point is that the syntax is also an instruction to the browser and may very well not be directly exposed to all users.

In my world, Lit-style templates would still use JS template literals with standard JS syntax, but be able to very easily use the browser's syntax to create an HTML template with parts:

```ts
const makeTemplate = (strings) => {
  const template = document.createElement('template');
  template.setAttribute('parseparts', '');
  template.innerHTML = strings.join('{{}}');
  return template;
```

The import part here is that the delimiter ('{{}}') is valid in attribute position, which rules out elements and comments.

bahrus:
@tbondwilkinson:

> The choice of {{}} is not the standard for all templating languages, but it's close enough that we can use it as a placeholder for now before we survey more developers.

I like the direction this is going.   If I'm understanding correctly, the most important feedback should come from the server-side generating frameworks.  If that is correct, I would encourage, if possible, surveying developers not just from the node-centric world, but include perhaps some representatives from [beyond](https://w3techs.com/technologies/overview/programming_language), a small slice of which may be using XSLT to generate the HTML.  It seems that the key will be how easy it would be for early adopter users initially to extend the template syntax to support the desired output.  Asp.net razor pages have support for custom tag helpers/ attribute helpers, which would allow for syntax that is defined in an elegant, semantic way (could even use xslt tags if so desired), which would emit the needed markers, similar to what it sounds like lit would support.  XSLT supports user defined functions which could also generate the desired output, I believe.

So it seems to me looking for syntax that matches the existing authoring / templating languages on the server side may not be as important as their ability to flexibly emit what is needed, but I might be missing something.  I suspect it may not matter all that much if {{}} is used versus ${}, for example, echoing what @justinfagnani  said.

proposals/DOM-Parts-Imperative.md
```diff
@@ -256,156 +216,87 @@ had initially contained `mailto:` before `{email}` but we could not capture this
...
 
 ```js
-const [firstName, lastName] = AttributePart.create(element, "title", null, [
-  null,
-  " ",
-  null,
-]);
-// Syntax to be improved. Here, a new AttributePart is created between each string.
+const part = AttributePart(document.getPartRoot(), element, "href");
+part.value = ["mailto: ", email];
```
rictic, line 226: I think we should think about making this behavior consistent and intuitive across the case where there are statics vs when there aren't. We also need to work with trusted types, so that we can set `part.value` to a trusted type in a page where trusted types are enforced.

One proposal:

An attribute is said to be fully controlled by an AttributePart's value if the statics are `['', '']`. In that case, commit behaves like:

```js
  #commitFullyControlled() {
    const value = Array.isArray(this.value) ? this.value[0] : this.value;
    if (value == null || value === false) {
      this.element.removeAttributeNS(this.namespaceURI, this.localName);
      return;
    }
    this.element.setAttributeNS(this.namespaceURI, this.localName, value);
  }
```

Note: if `value` is an array, we use its first element. This is to be consistent with the behavior when there are nonempty statics. `null`, `undefined`, and `false` when set on a fully controlled attribute remove it from the element.

If the attribute is not fully controlled by the value, then commit behaves like:

```js
  #commitWithStatics() {
    const value = Array.isArray(this.value) ? this.value : [this.value];
    let pieces = [this.statics[0]];
    for (let i = 1; i < statics.length; i++) {
      pieces.push(value[i - 1] ?? '', this.statics[i]);
    }
    return pieces.join('');
  }
```

Note: if the values array is too short, then the `?? ''` will write it as empty string, and excess elements are ignored.

And for completeness:

```js
  commit() {
    if (this.statics.length === 2 && this.statics[0] === '' && this.statics[1] === '') {
      this.#commitFullyControlled();
    } else {
      this.#commitWithStatics();
    }
  }
```

proposals/DOM-Parts-Imperative.md
```diff
@@ -256,156 +216,87 @@ had initially contained `mailto:` before `{email}` but we could not capture this
...
 
-- **Pros** No need for new `PartialAttributePart` or `AttributePart`
-  coordination. Does not block a future API object that represents partial
-  attributes.
-- **Cons** Need a way to represent partial attributes that are declaratively
-  defined.
+At this time there is no specific native representation of templates that could
+be inspected to determine with certainty which strings were compile time
+constants, but if some a representation were to exist, this API could provide
+better security for attributes that were sens
```
rictic, line 259: Looks like a bit was cut off in editing

proposals/DOM-Parts-Imperative.md
tbondwilkinson: I updated the wording a bit in this section to more cleanly spell out how statics and values fit together.

I'm not sure on statics = ['', ''] being the boundary condition. Why not just statics = undefined in the case where the AttributePart was created without statics? We can throw an Error during construction if someone tries to create an AttributePart with a static array that is less than 2 in length.

rictic: `['', '']` may be more elegant, but `undefined` may be more performant, and I'd prefer the more performant option. No objection from me for `undefined`

proposals/DOM-Parts-Declarative.md
```diff
@@ -0,0 +1,131 @@
...
+field.
+
+```
+{{# metadata1}}default value{{/ metadata2}}
+```
+
+## Choice of marker
+
+The `{{}}` and `{{#}}{{/}}` are reasonable DOM part markers, but this is open to
+proposals. It's possible to even allow the page to decide ewhat their markers
```
jatraman, line 130: nit: s/ewhat/what

proposals/DOM-Parts-Imperative.md
hax: I suppose the behavior should align with `String.raw`, so what's the logic difference between `#commitWithStatics` and `String.raw`? It seems the only difference is how to treat `null/undefined` in `this.value`. `String.raw` will treat them as `"null"` and `"undefined"`, but `#commitWithStatics` will use empty strings.

justinfagnani: `String.raw` is about how escape sequences are handled, which I don't think matters here.

PR #1062, opened 2024-07-17
Add referenceTarget to ShadowRootInit argument to attachShadow()

behowell:
From review feedback in the Chromium implementation CL, it may be good to add `referenceTarget` to the `ShadowRootInit` dictionary argument to `attachShadow()`.
https://chromium-review.googlesource.com/c/chromium/src/+/5615615/comment/1a3651e1_4b117a02/

### DIFF
--- a/proposals/reference-target-explainer.md
+++ b/proposals/reference-target-explainer.md
@@ -72,7 +72,10 @@ The goal of breaking it into phases is to get the simpler syntax and simpler use
 
 A component can specify an element in its shadow tree to act as its "reference target". When the host component is the target of a IDREF like a label's `for` attribute, the referenceTarget becomes the effective target of the label.
 
-The shadow root specifies the ID of the target element inside the shadow DOM. This is done either in JavaScript with the `referenceTarget` attribute on the `ShadowRoot` object, or in HTML markup using the `shadowrootreferencetarget` attribute on the `<template>` element.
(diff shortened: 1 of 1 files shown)


Westbrook (approved):
LGTM, quick question, however:

What happens in the time between setting the `referenceTarget`, via either technique, and that element actual existing due to the `innerHTML` being set? Should we talk about that in any way?

behowell:
@Westbrook 
> What happens in the time between setting the `referenceTarget`, via either technique, and that element actual existing due to the `innerHTML` being set? Should we talk about that in any way?

That's implicitly covered by the fact that reference target is a [Live Reference](https://github.com/WICG/webcomponents/blob/gh-pages/proposals/reference-target-explainer.md#live-references). Basically, the reference target gets updated any time something changes in the DOM that affects it. For example adding elements to the shadow DOM tree by setting `innerHTML` will update the reference target. It's equivalent to the non-shadow-dom case: if you have `<label for="foo">...</label>` and then _later_ add `<input id="foo">` to the DOM, the label will be "updated" to be for the input.

That said, it might be worth expanding the Live References section to include mentions of the DOM changing, instead of only mentioning the referenceTarget changing. I can make that change in a separate PR.

behowell:
Added a follow-up PR to clarify the live references: https://github.com/WICG/webcomponents/pull/1067

PR #1066, opened 2024-08-12
Reference Target: Clarify how JS Element attributes work with referenceTarget

behowell:
Update the **JavaScript attributes that reflect `Element` objects** to clarify which attributes return the host element. Also, add special cases for the `.form` and `.list` attributes, which return `null` because their types don't allow returning the host element.

### DIFF
--- a/proposals/reference-target-explainer.md
+++ b/proposals/reference-target-explainer.md
@@ -12,7 +12,7 @@ This proposal is based on [@Westbrook](https://github.com/Westbrook)'s [Cross-ro
 
 ### Cross-Root ARIA
 
-For an in-depth description the cross-root ARIA problem, see [@alice](https://github.com/alice)'s article [How Shadow DOM and accessibility are in conflict](https://alice.pages.igalia.com/blog/how-shadow-dom-and-accessibility-are-in-conflict/). The article describes the two main problems that need to be solved: 
+For an in-depth description the cross-root ARIA problem, see [@alice](https://github.com/alice)'s article [How Shadow DOM and accessibility are in conflict](https://alice.pages.igalia.com/blog/how-shadow-dom-and-accessibility-are-in-conflict/). The article describes the two main problems that need to be solved:
 
 #### 1. Referring from Shadow DOM outwards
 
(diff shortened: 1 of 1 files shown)


Westbrook (approved):
This explainer is getting closer and closer to spec text every time. Thanks for the added specificity!

proposals/reference-target-explainer.md
```diff
@@ -295,14 +295,27 @@ Reference target does not change the behavior of the host element when it is nes
...
+These will _never_ directly return the referenceTarget element that's inside the shadow tree. This is because an [IDL attribute with type Element](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#reflecting-content-attributes-in-idl-attributes:element) can only refer to an element that is a descendant of a [shadow-including ancestor](https://dom.spec.whatwg.org/#concept-shadow-including-ancestor) of the element hosting the attribute.
+
+Instead, most attributes return the **host** element that they're targeting, as long as the attribute's expected type is `HTMLElement`. However, The `.form` and `.list` attributes will return `null` when used with a referenceTarget, because they are expected to be `HTMLFormElement` or `HTMLDataListElement` and the host element itself is not a form or datalist. Importantly, the underlying association will still exist: the input will be connected to the form, for example; it's just not reflected by the `.form` attribute.
```
mfreed7, line 316: This feels a little odd, doesn't it? There **is** a form/list reference, and it's valid, but the JS accessor returns `null`. Perhaps it'd be better to have the IDL for `form` and `list` return `HTMLElement` (or even `Element`?)?

I suppose there are likely web compat issues with this, of some variety.

behowell: Yeah I agree it seems odd. My main concern was worrying not to break the web by changing the return type to `HTMLElement` instead of `HTMLFormElement`/`HTMLDataListElement`. That said, the change would only affect new scenarios that were previously not possible, and it should not affect existing code as far as I can tell. It would be good to have an expert weigh in on the potential problems with changing the IDL return types in the new scenarios.

Westbrook: Do you want to hold this PR for this feedback or open a new PR once we've heard from other implementors about possible issues here?

behowell: I think it would benefit from further discussion, but I don't think it should block this PR for now. I logged https://github.com/WICG/webcomponents/issues/1072 to discuss the follow-up options. Depending on the outcome of the discussion, another PR could be made to update this section.

Westbrook: Sounds good. Thanks Ben!

PR #1067, opened 2024-08-12
Reference Target: Clarify the Live References section

behowell:
Clarify in the Live References section that DOM changes also affect the reference target element.

This is a follow up to a PR comment https://github.com/WICG/webcomponents/pull/1062#pullrequestreview-2184344535.


Westbrook (approved):
Thanks for this clarification!

Is there a related section that points out what happens to references when `referenceTarget` doesn't resolve an element? Maybe the practicalities of "just like when there is no element of that ID" are assumed. However, the idea that there _is_ an element, but it's "nothing", in certain cases, could be surprising.

PR #1084, opened 2024-11-21
Update reference-target-explainer supported attributes

EisenbergEffect:
Just a small update to replace `invoketarget` with `commandfor` based on the latest invokers proposal.


Westbrook (approved):
LGTM!

